PR #1, opened 2019-05-06
Patches to the 0.6.0 release

clintfred:
* restored `UserCreateKeyPair` to public API
* created a binding module for binding specific code that needs to live in `ironoxide`
* added `IronOxideErr` to the `prelude`


src/binding.rs
```diff
@@ -0,0 +1,9 @@
+pub mod java {
+    use crate::IronOxideErr;
+    /// A way to turn IronSdkErr into Strings for the Java binding
+    impl From<IronOxideErr> for String {
```
coltfred: Since type class instances are found regardless of the modules they're in I think we should just put this in lib.rs. We're using it exclusively for our java binding, but others might not so putting it in a binding specific mod might not make sense to future consumers.

Alternatively we could refuse to provide this since it's generally a bad idea and make our code in ironoxide-java uglier ... This is a tough pill to swallow since it makes everything noisier, but would keep this code base cleaner.

clintfred: I agree that this doesn't enforce anything, but I thought it was a good hint that this code was in a file called binding. I don't see a downside to breaking it out for organization sake. I think part of the reason it got deleted in the first place was that it wasn't clear that it was needed outside of the crate. Maybe we'll also have other binding specific code?

It certainly could go in lib.rs. I think putting it in its own file is cleaner. We could also put some extra documentation on it in lib.rs.

coltfred: I disagree that putting it in a file called binding is making the organization better. It's not just applicable for binding sake. I hope we don't have other binding specific code and even this makes me a bit sad. Why it was deleted before likely has more to do with large refactoring and not caring too much about breaking the api going forward than some disregard for this impl. I'm not even sure when it was deleted... Was it you or @ernieturner who accidentally removed it?

If others feel that putting it in a binding file is good than I'll withdraw my objection, but as it stands now I don't see anything specific to binding about this implementation. We use it that way, but it's available for everyone regardless of if you're in a binding or not. If we don't want it available for others;a we should pursue some way of enforcing it, not trying to hide it off in a module (which will be included in the trait search path anyways).

PR #8, opened 2019-05-14
Add serde support to DeviceContext

clintfred:


src/internal/mod.rs
```diff
@@ -520,6 +560,30 @@ pub(crate) mod test {
...
+            TryFrom::try_from("account_id").unwrap(),
+            22,
+            priv_key.into(),
+            DeviceSigningKeyPair::from(dev_keys),
+        );
+
+        let json = serde_json::to_string(&context);
+
+        let de: DeviceContext = serde_json::from_str(&json.unwrap()).unwrap();
+        assert_eq!(context.account_id(), de.account_id());
```
ernieturner: It seems like asserting something that's a little more complex than just the account ID here might be good. The (de)serializing of the keys is what I think is more important to test.

src/internal/rest.rs
```diff
@@ -93,17 +93,23 @@ impl<'a> Authorization<'a> {
 }
 
 ///A struct which holds the basic info that will be needed for making requests to an ironcore service. Currently just the base_url.
-#[derive(Clone)]
-pub struct IronCoreRequest {
-    base_url: &'static str, //This will eventually be non-static.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct IronCoreRequest<'a> {
+    base_url: &'a str, //This will eventually be non-static.
```
ernieturner: Remove this comment?

src/internal/mod.rs
clintfred: Yep. Fair enough. It's annoying that we can't derive `PartialEq` because many of the underlying recrypt types don't support it. I am now asserting about both key types.

PR #9, opened 2019-05-22
Added test to show you can now share with a group you're not a member of.

coltfred:


coltfred:
Approved, since it's clint's PR.

PR #14, opened 2019-06-13
Add ability to encrypt without granting to self.

skeet70:


src/internal/document_api/mod.rs
```diff
@@ -417,36 +418,46 @@ pub fn encrypt_document<'a, CR: rand::CryptoRng + rand::RngCore>(
...
-                    user_errs,
-                    encrypt_errs.into_iter().map(|e| e.into()).collect(),
-                ]
-                .into_iter()
-                .concat();
-                (grants, encrypted_doc, other_errs)
-            })
+            let (group_errs, groups_with_key) = process_groups(groups);
+            let (user_errs, users_with_key) = process_users(users);
+            let self_grant = WithKey::new(
```
coltfred: This isn't useful if grant_to_author is false so I'd move it into the if.

src/internal/document_api/mod.rs
```diff
@@ -417,36 +418,46 @@ pub fn encrypt_document<'a, CR: rand::CryptoRng + rand::RngCore>(
...
+                },
+                user_master_pub_key.clone(),
+            );
+            let grant_list = if grant_to_author {
+                [&[self_grant], &users_with_key[..], &groups_with_key[..]].concat()
+            } else {
+                [&users_with_key[..], &groups_with_key[..]].concat()
+            };
+
+            if grant_list.is_empty() {
```
coltfred: I had to stare at this a while to make sense of it so it might be worth calling out that the reason you check this on the grant_list is because you would have already added self to the grant_list above if graunt_to_author was true.

src/document.rs
```diff
@@ -69,6 +76,7 @@ pub trait DocumentOps {
     /// - `encrypt_opts` - Optional document encrypt parameters. Includes
     ///       `id` - Unique ID to use for the document. Document ID will be stored unencrypted and must be unique per segment.
     ///       `name` - Non-unique name to use in the document. Document name will **not** be encrypted.
+    ///       `grant_to_author` - Flag determining whether to encrypt to the calling user or not.
```
ernieturner:
```suggestion
    ///       `grant_to_author` - Flag determining whether to encrypt to the calling user or not. If set to false at least one value must be present in the `grant` list.
```

src/internal/document_api/mod.rs
ernieturner: Maybe not a big deal, but it seems like maybe this validation ought to happen out at the call site in the `src/document.rs` file. We should be able to check that condition at that point right? I can't remember if we have any convention on this, but it seems like making sure that the arguments are valid before we call into the various `mod.rs` files might be good.

tests/document_ops.rs
```diff
@@ -72,6 +74,68 @@ fn doc_create_with_grant() {
...
+    let doc_result = sdk.document_encrypt(
+        &doc,
+        &DocumentEncryptOpts::new(None, Some("first name".try_into().unwrap()), false, vec![]),
+    );
+
+    assert_eq!(
+        doc_result.err().unwrap(),
+        IronOxideErr::ValidationError(
+            "grant_to_author".to_string(),
+            "grant_to_author cannot be false if there are no explicit grants".to_string()
```
ernieturner: I'm not a huge fan of duplicating this exact string in the unit tests. I think it'd be good to have some verification (that it contains "grant_to_author" or similar). But I don't think we care enough about this specific error string that if we decide to change it the unit tests should then start failing.

skeet70: Is this better?

```rs
    // make sure there was a validation error, and that the problem was with the grant
    assert_eq!(
        match doc_result.err().unwrap() {
            IronOxideErr::ValidationError(field_name, _) => field_name,
            _ => "failed test".to_string(),
        },
        "grant_to_author".to_string()
    )
```

ernieturner: 👍

Issue #15, opened 2019-06-26
Add sdk.document_encrypt_via_policy


coltfred:
The service now supports the ability to use policy to determine who to encrypt to. In order to support this in IronOxide, we should add a function `document_encrypt_via_policy` which takes all the same options as `document_encrypt`, but instead of `grants` it takes 4 new options.

`classification: Option<String>`, `sensitivity: Option<String>`, `data_subject: Option<String>` `substitute_id: Option<String>` each of these do have limited character sets so if we want we could make newtypes that valid by construction.

The first 3 must follow the following regex: `[A-Za-z0-9_-]{1,100}`. The last one must be a valid user id in the IronCore service so taking `User` there instead of String would be preferable.

The api that should be called is `GET /api/1/policys?classification=<value>&sensitivity=<value2>&dataSubject=<value3>&substitutionId=<id>`. All values are optional and should not be included if they were None.

The return structure will be the following:
```
struct PolicyResult{
  users: List<IdAndKey>,
  groups: List<IdAndKey>,
  invalidUsers: List<ProvidedId>,
  invalidGroups: List<ProvidedId>
}

struct IdAndKey{
  id: ProvidedId,
  key: PublicKey
}
```

The invalidUsers and invalidGroups will be filled out if the policy resulted an ids that could not be found in the ironcore system, if they are empty all groups and users were successfully looked up.

PR #16, opened 2019-07-09
[#15] Encrypt via policy

clintfred:
Encrypt a document via a server side policy!

This feature can be used to take the burden off the caller to assemble the correct set of users and groups that need access to the document. A policy is identified by the triple (Category, Sensitivity, Data Subject) and is stored in the IronCore web service. Given values for the triple, the policy can be evaluated to get the correct set of users and groups from the web service during an encrypt operation.

Policy-based encryption can also be combined with explicit grants.

Read more about policy in policy.rs

## TODO prior to merge
* [x] - additional documentation added to sdk (and this PR)
* [x] - consider moving policy related code to it's own module
* [x] - add a couple more integration tests

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -3,13 +3,13 @@ pub use crate::internal::document_api::{
     DocumentEncryptResult, DocumentListMeta, DocumentListResult, DocumentMetadataResult,
     UserOrGroup, VisibleGroup, VisibleUser,
 };
-use crate::policy::*;
 use crate::{
     internal::{
         document_api::{self, DocumentId, DocumentName},
         group_api::GroupId,
         user_api::UserId,
     },
+    policy::*,
     Result,
 };
 use itertools::{Either, EitherOrBoth, Itertools};
--- a/src/document.rs
+++ b/src/document.rs
@@ -3,7 +3,7 @@ pub use crate::internal::document_api::{
     DocumentEncryptResult, DocumentListMeta, DocumentListResult, DocumentMetadataResult,
     UserOrGroup, VisibleGroup, VisibleUser,
 };
-use crate::internal::validate_simple_policy_id;
+use crate::internal::validate_simple_policy_field_id;
 use crate::{
     internal::{
         document_api::{self, DocumentId, DocumentName},
@@ -77,12 +77,77 @@ impl<'a> DocumentEncryptOpts {
 
 /// Document access granted by a policy.
 ///
-/// A policy is stored on the server and yields a list of users and groups when applied.
+/// Policies are defined using the ironcore admin console: https://admin.ironcorelabs.com/policy
+/// In the future policies will be able to be programmatically defined using ironoxide.
 ///
-/// `category` -
-/// `sensitivity` -
-/// `data_subject` -
-/// `substitute_id` -
+/// A policy could look something like:
+/// ```json
+/// {
+///  "dataSubjects": [
+///    "PATIENT",
+///    "EMPLOYEE"
+///  ],
+///  "sensitivities": [
+///    "RESTRICTED",
+///    "CLASSIFIED",
+///    "INTERNAL"
+///  ],
+///  "categories": [
+///    "HEALTH",
+///    "PII"
+///  ],
+///  "rules": [
+///    {
+///      "sensitivity": "RESTRICTED",
+///      "users": [
+///        "%USER%"
+///      ],
+///      "dataSubject": "PATIENT",
+///      "groups": [
+///        "group_other_%USER%",
+///        "group_id_doctors",
+///        "data_recovery"
+///      ],
+///      "category": "HEALTH"
+///    },
+///    {
+///      "sensitivity": "INTERNAL",
+///      "users": [
+///        "joe@ironcorelabs",
+///        "%LOGGED_IN_USER%"
+///      ],
+///      "groups": [
+///        "group_%LOGGED_IN_USER%",
+///        "data_recovery"
+///      ],
+///      "category": "PII"
+///    },
+///    {
+///      "groups": [
+///        "data_recovery"
+///      ],
+///    },
+///  ]
+/// }
+/// ```
+///
+/// A policy is stored on the server and yields a list of users and groups when applied.
+/// The triple (`category`, `sensitivity`, `data_subject`) maps to a single policy rule. Each policy
+/// rule may generate any number of users/groups.
+///
+/// How a `substitute_user_id` is a `UserId` that will be used to replace the `%USER%` placeholder.
+/// `%LOGGED_IN_USER%` is a special token that will be replaced by the user currently authenticated
+/// to make SDK calls.
+///
+/// Rules in the policy are matched from top to bottom. If more than one rule would match a `PolicyGrant`
+/// the first one will be applied.
+///
+/// Example:
+/// If the current user of the sdk is "alice@ironcorelabs" and the following PolicyGrant is evaluated,
+/// `PolicyGrant::new("PII".try_from()?, "INTERNAL".try_from()?, None, None)` will match the last rule
+/// in the example policy, above and will return users: [joe@ironcorelabs, alice@ironcorelabs] and
+/// groups [group_alice@ironcorelabs, data_recovery"]
+///
 #[derive(Debug, PartialEq, Clone)]
 pub struct PolicyGrant {
     category: Option<Category>,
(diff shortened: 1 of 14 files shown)


src/document.rs
```diff
@@ -33,14 +55,139 @@ impl<'a> DocumentEncryptOpts {
...
+            grants: EitherOrBoth::Right(policy),
+        }
+    }
+}
+
+/// Document access granted by a policy.
+///
+/// A policy is stored on the server and yields a list of users and groups when applied.
+///
+/// `category` -
```
coltfred: Comments need finishing.

src/document.rs
```diff
@@ -33,14 +55,139 @@ impl<'a> DocumentEncryptOpts {
...
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct Category(pub(crate) String);
+
+impl TryFrom<&str> for Category {
+    type Error = IronOxideErr;
+
+    fn try_from(value: &str) -> Result<Self> {
+        validate_simple_policy_id(value, "Category").map(|v| Self(v))
```
coltfred: You can replace this with `.map(Self::new)`, which I personally like better because it doesn't introduce a variable in scope.

src/document.rs
```diff
@@ -33,14 +55,139 @@ impl<'a> DocumentEncryptOpts {
...
+}
+
+impl Category {
+    pub(crate) const QUERY_PARAM: &'static str = "category";
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct Sensitivity(pub(crate) String);
+
+//TODO macro?
```
coltfred: Meh, I'd just remove the comment and move on.

src/document.rs
```diff
@@ -33,14 +55,139 @@ impl<'a> DocumentEncryptOpts {
...
+impl TryFrom<&str> for Sensitivity {
+    type Error = IronOxideErr;
+
+    fn try_from(value: &str) -> Result<Self> {
+        validate_simple_policy_id(value, "Sensitivity").map(|v| Self(v))
+    }
+}
+
+impl Sensitivity {
+    pub(crate) const QUERY_PARAM: &'static str = "sensitivity";
```
coltfred: Can you make this a const string and borrow it when you need?

src/document.rs
```diff
@@ -84,6 +231,21 @@ pub trait DocumentOps {
         encrypt_opts: &DocumentEncryptOpts,
     ) -> Result<DocumentEncryptResult>;
 
+    //    /// Encrypt the provided document bytes using the specified policy.
+    //    ///
+    //    /// # Arguments
```
coltfred: Looks like you don't want this?

src/internal/document_api/mod.rs
```diff
@@ -685,6 +738,95 @@ fn process_users(
...
+                                    group,
+                                    format!("Error parsing group public key {:?}", &key),
+                                )
+                            }),
+                    )
+                }
+
+                any => Either::Left(DocAccessEditErr::new(
+                    any.clone().into(),
+                    "User or group does not have associated public key".to_string(),
```
coltfred: I know it's a bit more verbose, but the caller won't know which one didn't have a key. Is it a user or is it a group?

src/internal/document_api/mod.rs
```diff
@@ -685,6 +736,95 @@ fn process_users(
     (user_errs, users_with_key)
 }
 
+fn process_policy(
```
ernieturner: Ditto here

src/internal/document_api/mod.rs
```diff
@@ -685,6 +736,95 @@ fn process_users(
...
+    policy_result: &PolicyResult,
+) -> (Vec<DocAccessEditErr>, Vec<WithKey<UserOrGroup>>) {
+    let (de_errs, policy_eval_results): (Vec<DocAccessEditErr>, Vec<WithKey<UserOrGroup>>) =
```
ernieturner: I'm not sure it's obvious what `de` stands for here. Something like `doc_edit_errs` is still pretty terse but easier to parse

src/internal/document_api/requests.rs
```diff
@@ -213,6 +226,58 @@ pub mod document_create {
...
+        }
+    }
+
+    use crate::document::{Category, DataSubject, PolicyGrant, Sensitivity, SubstituteId};
+
+    pub fn policy_get_request(
+        auth: &RequestAuth,
+        policy_grant: &PolicyGrant,
+    ) -> impl Future<Item = PolicyResult, Error = IronOxideErr> {
+        dbg!(&policy_grant);
```
ernieturner: Remove

src/internal/mod.rs
```diff
@@ -102,6 +103,9 @@ quick_error! {
         NotGroupAdmin(group_id:group_api::GroupId) {
             display("You're are not an administrator of group '{}'", group_id.0)
         }
+        InvalidPolicy {
+            display("A policy must specify one or more of category, sensitivity, data_subject, substitute_id")
```
ernieturner: Pretty minor, but I'd remove the underscores from these names in this error

src/internal/mod.rs
```diff
@@ -147,6 +151,24 @@ pub fn validate_id(id: &str, id_type: &str) -> Result<String, IronOxideErr> {
     }
 }
 
+pub fn validate_simple_policy_id(id: &str, id_type: &str) -> Result<String, IronOxideErr> {
```
ernieturner: Maybe `validate_simple_policy_field(field, field_type)` or something? They aren't really IDs in the way we usually talk about them.

src/internal/mod.rs
```diff
@@ -147,6 +151,24 @@ pub fn validate_id(id: &str, id_type: &str) -> Result<String, IronOxideErr> {
...
+    let simple_policy_field_regex = Regex::new("^[A-Za-z0-9_-]+$").expect("regex is valid");
```
ernieturner: Ya, like the name you're using here 😄

src/internal/rest.rs
```diff
@@ -203,66 +228,61 @@ impl<'a> IronCoreRequest<'a> {
...
         builder = maybe_body
             .iter()
             .fold(builder, |build, body| build.json(body));
 
-        builder
-            .headers(DEFAULT_HEADERS.clone())
-            .headers(headers)
-            .send()
+        let req = builder.headers(DEFAULT_HEADERS.clone()).headers(headers);
+        dbg!(&req);
```
ernieturner: Remove

src/internal/rest.rs
```diff
@@ -203,66 +228,61 @@ impl<'a> IronCoreRequest<'a> {
...
+        req.send()
             //Parse the body content into bytes
             .and_then(|res| {
+                dbg!(&res);
```
ernieturner: Remove

src/internal/rest.rs
```diff
@@ -203,66 +228,61 @@ impl<'a> IronCoreRequest<'a> {
...
                 let status_code = res.status();
-                res.into_body()
-                    .concat2()
-                    .map(move |body| (status_code, body))
+                res.into_body().concat2().map(move |body| {
+                    dbg!(&body);
```
ernieturner: Remove

src/document.rs
clintfred: I don't think this is true, at least not without additional traits implemented.

src/internal/document_api/mod.rs
clintfred: `de` is common in Rust for deserialization. The errors are the result of public keys that failed to deserialize. I'll rename to `pubkey_errs` and see if that makes more sense.

src/internal/document_api/mod.rs
clintfred: True. We have a nice Display for UserOrGroup, so I reworked it to use that.

src/internal/mod.rs
clintfred: I removed the whole error as it's not needed.

src/internal/mod.rs
clintfred: `validate_simple_policy_field_id` seem okay?

ernieturner: Somewhat, sure. I'd still prefer to remove the `_id` there, but I'm not going to argue if you want to keep it.

clintfred:
I still need to sweep though and remove `dbg` calls. I was thinking of putting the policy stuff in it's own module, but not sure on that yet. I will read the policy documentation I wrote again to see if I can improve it further.

src/document.rs
clintfred: I'm not sure what you're suggesting. These are const, but maybe you are hoping for something else?

src/document.rs
clintfred: Macro added as it was easy.

clintfred:
@ernieturner I did upgrade Travis (and my local env) to Rust 1.36.0. No issues.

src/policy.rs
```diff
@@ -0,0 +1,263 @@
+//! Policies are a list of rules which map data labels to a list of users/groups. This allows the
+//! separation of concerns when it comes to labeling data vs defining who to encrypt to.
+//!
+//! Policies are defined using the ironcore admin console: https://admin.ironcorelabs.com/policy
+//! and are stored on the server. This allows a policy to be updated independently any application code.
```
cjyar: "updated independently OF any application code"

src/policy.rs
```diff
@@ -0,0 +1,263 @@
...
+//! In the future policies will be able to be programmatically defined using ironoxide.
+//!
+//! Data labeling is provided in three dimensions (category, sensitivity, dataSubject).
+//! You only need to use the dimensions that make sense for your use case. The values of the labels
+//! are arbitrary, but the example below may be instructive in selecting label names.
+//!
+//! In addition to defining labels, a list of rules is required to map the labels to a set of users/groups.
+//! Rules are checked in the order they are defined. If a rule matches is can produce any number of users/groups.
```
cjyar: s/is/it/

PR #17, opened 2019-07-12
Remove type SubstituteId 

clintfred:
...and make string values of PolicyGrant components available to the outside world

### DIFF
--- a/src/internal/user_api/mod.rs
+++ b/src/internal/user_api/mod.rs
@@ -19,7 +19,7 @@ mod requests;
 #[derive(PartialEq, Debug, Serialize, Deserialize, Clone, Eq, Hash)]
 pub struct UserId(pub(crate) String);
 impl UserId {
-    pub fn id(&self) -> &String {
+    pub fn id(&self) -> &str {
         &self.0
     }
 
--- a/src/policy.rs
+++ b/src/policy.rs
@@ -84,13 +84,13 @@ use std::convert::TryFrom;
 /// The triple (`category`, `sensitivity`, `data_subject`) maps to a single policy rule. Each policy
 /// rule may generate any number of users/groups.
 ///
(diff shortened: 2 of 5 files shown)


src/policy.rs
```diff
@@ -84,13 +84,13 @@ use std::convert::TryFrom;
 /// The triple (`category`, `sensitivity`, `data_subject`) maps to a single policy rule. Each policy
 /// rule may generate any number of users/groups.
 ///
-/// `substitute_user_id` replaces `%USER%` in a matched policy rule.
+/// `substitute_user_id replaces `%USER%` in a matched policy rule.
```
ernieturner: Did you mean to remove this backtick?

src/internal/user_api/mod.rs
```diff
@@ -19,7 +19,7 @@ mod requests;
 #[derive(PartialEq, Debug, Serialize, Deserialize, Clone, Eq, Hash)]
 pub struct UserId(pub(crate) String);
 impl UserId {
-    pub fn id(&self) -> &String {
+    pub fn id(&self) -> &str {
```
ernieturner: Is this considered a breaking change? This struct gets exposed externally, right?

src/policy.rs
clintfred: good catch.

src/internal/user_api/mod.rs
clintfred: Yes it is part of our public API and is technically a breaking change. For most (all?) cases `&str` is more idiomatic than `&String` (similar to `&Option<T>` vs `Option<&T>`). If it's actually a breaking change depends on on the usage. Rust provides From and Deref impls that mitigate this. There was one change required to ironoxide-java (`.clone()` -> `.to_string()`)

We undoubtedly have more of these `&String` instances, and I could certainly see the sense in changing them all at once vs piecemeal.

PR #26, opened 2019-07-31
Update to ring 0.16

giarc3:
Updating ring dependency to 0.16 and lazy_static to 1.3
Wrote NonceSequence wrapper for SealingKey/OpeningKey
Fixed author email typo

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "ironoxide"
 version = "0.9.0"
-authors = ["IronCore Labs <info at ironcorelabs.com>"]
+authors = ["IronCore Labs <info@ironcorelabs.com>"]
 readme = "README.md"
 license = "AGPL-3.0-only"
 repository = "https://github.com/IronCoreLabs/ironoxide"
@@ -20,15 +20,15 @@ serde_derive = "~1.0"
 rand = "~0.6"
 rand_chacha = "~0.1"
 regex = "~1.1"
-ring = "~0.14"
-recrypt = "~0.8"
+ring = "~0.16"
+recrypt = "0.8.1"
 reqwest = "~0.9"
 tokio = "~0.1"
 hex = "~0.3"
 itertools = "~0.8"
 futures = "~0.1.25"
 quick-error = "~1.2"
-lazy_static = "~1.2"
+lazy_static = "~1.3"
 chrono = { version = "0.4", features = ["serde"] }
 percent-encoding="~1.0"
 log = "~0.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ rand = "~0.6"
 rand_chacha = "~0.1"
 regex = "~1.1"
 ring = { version= "~0.16", features = ["std"] }
-recrypt = "0.8.1"
+recrypt = "~0.8.1"
 reqwest = "~0.9"
 tokio = "~0.1"
 hex = "~0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ serde_derive = "~1.0"
 rand = "~0.6"
 rand_chacha = "~0.1"
 regex = "~1.1"
-ring = "~0.16"
+ring = { version= "~0.16", features = ["std"] }
 recrypt = "0.8.1"
 reqwest = "~0.9"
 tokio = "~0.1"
--- a/src/crypto/aes.rs
+++ b/src/crypto/aes.rs
@@ -114,7 +114,7 @@ impl TryFrom<&[u8]> for AesEncryptedValue {
 }
 
 impl From<ring::error::Unspecified> for IronOxideErr {
-    fn from(ring_err: Unspecified) -> Self {
+    fn from(ring_err: ring::error::Unspecified) -> Self {
         IronOxideErr::AesError(ring_err)
     }
 }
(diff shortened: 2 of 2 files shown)


src/crypto/aes.rs
```diff
@@ -226,15 +229,14 @@ pub fn decrypt(
...
-    let aes_key = aead::OpeningKey::new(&aead::AES_256_GCM, &key[..])?;
-
-    let plaintext = aead::open_in_place(
-        &aes_key,
-        aead::Nonce::assume_unique_for_key(encrypted_doc.aes_iv),
-        aead::Aad::empty(),
-        0,
-        &mut encrypted_doc.ciphertext[..],
-    )?;
+    let unbound_key = aead::UnboundKey::new(&aead::AES_256_GCM, &key[..])?;
```
coltfred: Ditto here.

src/crypto/aes.rs
```diff
@@ -175,6 +176,16 @@ pub fn decrypt_user_master_key(
     Ok(fixed_decrypted_master_key)
 }
 
+struct NonceGenerator {
```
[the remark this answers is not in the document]
giarc3: Done

src/crypto/aes.rs
```diff
@@ -175,6 +176,16 @@ pub fn decrypt_user_master_key(
...
+    iv: [u8; aead::NONCE_LEN],
+}
+
+impl aead::NonceSequence for NonceGenerator {
+    fn advance(&mut self) -> Result<aead::Nonce, Unspecified> {
+        aead::Nonce::try_assume_unique_for_key(&self.iv)
```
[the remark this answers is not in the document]
giarc3: Done

src/crypto/aes.rs
```diff
@@ -184,20 +195,12 @@ pub fn encrypt<R: CryptoRng + RngCore>(
 ) -> Result<AesEncryptedValue, Unspecified> {
     let algorithm = &aead::AES_256_GCM;
     let mut iv = [0u8; aead::NONCE_LEN];
-
     take_lock(rng).deref_mut().fill_bytes(&mut iv);
-    let aes_key = aead::SealingKey::new(algorithm, &key[..])?;
-
+    let unbound_key = aead::UnboundKey::new(algorithm, &key[..])?;
+    let mut aes_key = aead::SealingKey::new(unbound_key, NonceGenerator { iv });
```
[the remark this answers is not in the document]
giarc3: Done

src/crypto/aes.rs
giarc3: Done

src/crypto/aes.rs
```diff
@@ -175,6 +176,16 @@ pub fn decrypt_user_master_key(
     Ok(fixed_decrypted_master_key)
 }
 
+struct ConstantNonceGenerator {
```
clintfred: Another naming suggestion: `SingleNonceGenerator`

I think this should also have a comment saying something about how this is safe because we only use it once and we generate the nonce from a CSPRG.

src/crypto/aes.rs
```diff
@@ -175,6 +176,16 @@ pub fn decrypt_user_master_key(
...
+    iv: [u8; aead::NONCE_LEN],
+}
+
+impl aead::NonceSequence for ConstantNonceGenerator {
+    fn advance(&mut self) -> Result<aead::Nonce, Unspecified> {
+        Ok(aead::Nonce::assume_unique_for_key(self.iv))
```
clintfred: If we wanted to be a little more explicit, we could keep a flag for if the generator had been previously used and return an `Err` in that case. (assuming my understanding of the usage of this struct is correct)

src/crypto/aes.rs
```diff
@@ -124,7 +125,7 @@ impl From<ring::error::Unspecified> for IronOxideErr {
 fn derive_key_from_password(password: &str, salt: [u8; PBKDF2_SALT_LEN]) -> [u8; AES_KEY_LEN] {
     let mut derived_key = [0u8; digest::SHA256_OUTPUT_LEN];
     pbkdf2::derive(
-        &digest::SHA256,
+        pbkdf2::PBKDF2_HMAC_SHA256,
```
clintfred: This is equivalent to the old value, correct?

Cargo.toml
```diff
@@ -20,15 +20,15 @@ serde_derive = "~1.0"
 rand = "~0.6"
 rand_chacha = "~0.1"
 regex = "~1.1"
-ring = "~0.14"
-recrypt = "~0.8"
+ring = { version= "~0.16", features = ["std"] }
+recrypt = "~0.8.1"
```
clintfred: This change is okay, but should have the same meaning as before.

src/crypto/aes.rs
giarc3: I expect so, but the old value gave an error of
`expected struct ring::pbkdf2::Algorithm, found struct ring::digest::Algorithm`
so I thought that this was what it wanted now

Cargo.toml
giarc3: This is how @coltfred helped me fix the problem with the AesError

src/crypto/aes.rs
clintfred: Looks right to me.

Cargo.toml
clintfred: That shouldn't have had any effect, but maybe there's something with caching in cargo that I don't understand.

coltfred: @clintfred The effect it has is that it ensures it will be at least 0.8.1 instead of at least 0.8.0. You're right that it doesn't really matter since it would always choose the latest anyways.

Also, Craig thought you were talking about the ring version line.

Issue #21, opened 2019-07-17
Support encrypt returning EDEKs instead of storing them with IronCore


coltfred:
This api should be `document_edek_encrypt(data: &[u8], encrypt_opts:EncryptOpts): (DocumentEncryptResult, Edeks)` (Not sure about tuple vs newtype).

This api should generate the dek and encrypt using the normal workflow. Instead of sending that as a document create to the api, it should instead take all that data and encode it using the protobuf schema that we've defined. Return that directly to the user.

This is the first api where we'll be working with PB as well as declaring a new trait and refactoring the current encrypt code.

PR #27, opened 2019-08-01
[#21] Implement Document Encryption that returns EDEKs

clintfred:
see #21 

## Changelog
* `DocumentAdvancedOps::document_encrypt_unmanaged` function added for advanced use cases where the calling application wants to manage both the encrypted data and the associated edeks instead of using the IronCore service for EDEK management.

## TODO
* [x] - sync up with latest edek proto definition
* [x] - ~fix android-64 build~ (disabled for now)
* [x] - convert remaining TODOs to other issues


clintfred:
Android 64 build is failing as it is cross compiled and doesn't allow writing to the `src/` tree. This is problematic as our build.rs generates `edeks.rs` at build time.

https://github.com/rust-embedded/cross/issues/145#issuecomment-409366182 may contain some helpful information if we decide to continue generating proto -> rs at ironoxide build time. Deferring this until we decide on a general way to handle .proto files.

build.rs
```diff
@@ -0,0 +1,15 @@
+extern crate protobuf_codegen_pure;
+
+fn main() {
+    protobuf_codegen_pure::run(protobuf_codegen_pure::Args {
+        out_dir: "src/proto",
+        input: &["proto/edeks.proto"],
+        includes: &["proto"],
+        customize: protobuf_codegen_pure::Customize {
+            carllerche_bytes_for_bytes: Some(true),
```
BobWall23: He needs to put his name in the functions he writes?? Sigh.

src/internal/document_api/mod.rs
```diff
@@ -522,28 +708,148 @@ fn encrypt_document_core<'a, CR: rand::CryptoRng + rand::RngCore>(
...
+        let mut proto_edek_data_no_pubk = match edek.encrypted_dek_data {
+            re::EncryptedValue::EncryptedOnceValue {
+                ephemeral_public_key,
+                encrypted_message,
+                auth_hash,
+                public_signing_key,
+                signature,
+            } => {
+                let mut proto_edek_data = EncryptedDekDataP::default();
+                {
```
coltfred: I was confused about why this block existed, but realize now that you're just using it to hide the mutable ephemeral public key. 

I think it'd be more clear if you assigned the immutable result to a variable and did the `set_ephemeralPublicKey` outside.

src/internal/document_api/mod.rs
```diff
@@ -522,28 +708,148 @@ fn encrypt_document_core<'a, CR: rand::CryptoRng + rand::RngCore>(
...
+                }
+                proto_edek_data.set_encryptedMessage(encrypted_message.bytes()[..].into());
+                proto_edek_data.set_signature(signature.bytes()[..].into());
+                proto_edek_data.set_authHash(auth_hash.bytes()[..].into());
+                proto_edek_data.set_publicSigningKey(public_signing_key.bytes()[..].into());
+
+                proto_edek_data
+            }
+            re::EncryptedValue::TransformedValue { .. } => {
+                unreachable!("Will be needed for decrypt!")
```
coltfred: I don't think this is possible with our current proto definition, though we could certainly allow it. I also don't think this is needed for decrypt.

src/internal/document_api/mod.rs
```diff
@@ -522,28 +708,148 @@ fn encrypt_document_core<'a, CR: rand::CryptoRng + rand::RngCore>(
...
+
+                let mut proto_pub_key = edeks::PublicKey::default();
+                proto_pub_key.set_x(public_key.as_bytes().into());
+                proto_pub_key.set_y(public_key.as_bytes().into());
+                (proto_uog, proto_pub_key)
+            }
+        };
+
+        // attach the ephemeral public key
+        proto_edek_data_no_pubk.set_ephemeralPublicKey(proto_pub_key);
```
coltfred: I don't think this is right. Why are you setting the ephemeral public key to a public key you pulled off a user or group?

src/internal/document_api/mod.rs
```diff
@@ -1054,4 +1362,31 @@ mod tests {
         let deduplicated_grants = dedupe_grants(&grants_w_dupes);
         assert_that!(&deduplicated_grants.len(), eq(2))
     }
+
+    #[test]
+    fn roundtrip_encrypted_dek_proto() {
```
coltfred: This test doesn't assert anything about the code you wrote above. Round trip would work even if you put everything in the wrong place. 

We need a test that asserts the translation is correct.

src/internal/document_api/mod.rs
```diff
@@ -1054,4 +1362,31 @@ mod tests {
...
+                public_key: pubk.into(),
+                id: UserId::unsafe_from_string("userid".to_string())
+                    .borrow()
+                    .into(),
+            },
+        };
+
+        let _proto_edek: EncryptedDekP = edek.borrow().into();
+
+        //TODO write the other half of this test for decrypt
```
coltfred: I don't think this test should have anything to do with decrypt.

src/internal/document_api/requests.rs
```diff
@@ -63,12 +65,12 @@ pub struct AccessGrant {
     pub(crate) encrypted_value: EncryptedOnceValue,
 }
 
+//TODO get rid of
```
coltfred: ?

src/proto/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod edeks;
```
coltfred: I think we decided on transform as the module name.

src/internal/document_api/mod.rs
```diff
@@ -435,53 +500,177 @@ pub fn encrypt_document<
     grant_to_author: bool,
     user_grants: &'a Vec<UserId>,
     group_grants: &'a Vec<GroupId>,
-    policy_grant: Option<&PolicyGrant>,
+    policy_grant: Option<&'a PolicyGrant>,
 ) -> impl Future<Item = DocumentEncryptResult, Error = IronOxideErr> + 'a {
     let (dek, doc_sym_key) = transform::generate_new_doc_key(recrypt);
```
BobWall23: A comment here might be nice - I assume `dek` is the `Plaintext` and `doc_sym_key` is the sha256 hash of it?

src/internal/document_api/mod.rs
```diff
@@ -435,53 +500,177 @@ pub fn encrypt_document<
...
+        ))
+        .and_then(move |(encrypted_doc, (grants, key_errs))| {
+            encrypt_document_core(
+                auth,
+                recrypt,
+                dek,
+                encrypted_doc,
+                &doc_id,
+                grants(user_master_pub_key),
+            )
```
BobWall23: This encrypts to the user master key even if you don't want to encrypt the document to the author?

src/internal/document_api/mod.rs
```diff
@@ -435,53 +500,177 @@ pub fn encrypt_document<
...
+            .into_future()
+            .and_then(move |r| {
+                document_create(
+                    &auth,
+                    r.clone(),
```
BobWall23: Is this cloning the encrypted document bytes?

src/internal/document_api/mod.rs
```diff
@@ -435,53 +500,177 @@ pub fn encrypt_document<
...
+    aes::encrypt_future(rng, &plaintext.to_vec(), *doc_sym_key.bytes())
+        .join(resolve_keys_for_grants(
+            auth,
+            user_grants,
+            group_grants,
+            policy_grant,
+            grant_to_author,
+        ))
+        .and_then(move |(encrypted_doc, (grants, key_errs))| {
+            encrypt_document_core(
```
BobWall23: Didn't `encrypt_future` actually encrypt the document?

src/internal/document_api/mod.rs
```diff
@@ -435,53 +500,177 @@ pub fn encrypt_document<
...
+                    }
+                },
+                [group_errs, user_errs, policy_errs].concat(),
             )
         })
 }
 
+/// Encrypts a document but does not create the document in the IronCore system.
+/// The resultant DocumentDetachedEncryptResult both the EncryptedDeks and the AesEncryptedValue for the caller to deal with.
+pub fn edek_encrypt_document<'a, R1, R2: 'a>(
```
BobWall23: seems a lot of shared code between this and `encrypt_document` - could the common code be factored out into a new function?

src/internal/document_api/mod.rs
```diff
@@ -522,28 +708,148 @@ fn encrypt_document_core<'a, CR: rand::CryptoRng + rand::RngCore>(
...
+    encrypted_dek_data: recrypt::api::EncryptedValue,
+}
+
+impl From<&EncryptedDek> for EncryptedDekP {
+    fn from(edek: &EncryptedDek) -> Self {
+        use crate::proto::edeks;
+        use recrypt::api as re;
+        let edek = edek.clone();
+
+        // transform the recrypt EncryptedValue to a edek proto
```
BobWall23: `transform` is a little confusing - I was thinking transform crypto when I saw it. Maybe `encode the EncryptedValue as a protobuf`?

src/internal/document_api/mod.rs
```diff
@@ -522,28 +708,148 @@ fn encrypt_document_core<'a, CR: rand::CryptoRng + rand::RngCore>(
...
+                    UserOrGroup::User {
+                        id: UserId(user_string),
+                    },
+                public_key,
+            } => {
+                let mut proto_uog = edeks::UserOrGroup::default();
+                proto_uog.set_userId(user_string.into());
+
+                let mut proto_pub_key = edeks::PublicKey::default();
+                proto_pub_key.set_x(public_key.as_bytes().into());
```
BobWall23: Maintaining separate `x` and `y` is seeming less and less appealing. I think we've discussed treating the public key as a black box - any further thoughts?

src/internal/document_api/mod.rs
```diff
@@ -250,6 +254,68 @@ impl DocumentMetadataResult {
...
+    pub fn encrypted_deks(&self) -> &[u8] {
+        &self.encrypted_deks
+    }
+    pub fn access_errs(&self) -> &[DocAccessEditErr] {
+        &self.access_errs
+    }
+
+    /// Users and Groups that have access to the encrypted document.
+    ///
+    /// Implementation is somewhat expensive as it is decoding the `encrypted_deks`.
```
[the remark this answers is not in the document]
clintfred: My thinking was that compared to all other `grants()` (`succeeded()` on `DocumentAccessResult) implementations we have, this isn't just returning `struct` data, but  is doing computation. I was trying to present the same interface as normal encrypt, but perhaps this function should have a different name, like `compute_grants` to communicate to the caller that this isn't a "accessor-type" function. 

Or maybe it doesn't matter...

src/internal/document_api/mod.rs
```diff
@@ -250,6 +254,68 @@ impl DocumentMetadataResult {
...
+    pub fn grants(&self) -> Vec<UserOrGroup> {
+        let pb_result: ProtobufResult<EncryptedDeksP> =
+            protobuf::parse_from_bytes(&self.encrypted_deks);
+        let result: Vec<UserOrGroup> = pb_result
+            // safe since the proto-encoded data is encoded in this file using the same definition as this decode
+            .expect("Unable to decode protobuf encoded encrypted_data")
```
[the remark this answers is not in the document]
clintfred: I don't understand this. `DocumentDetachedEncryptResult` can only be created in the `document_api` module. As long as that remains true, I think expecting the bytes to be parse-able is reasonable. Or maybe I'm missing something here?

src/internal/document_api/mod.rs
```diff
@@ -522,28 +708,148 @@ fn encrypt_document_core<'a, CR: rand::CryptoRng + rand::RngCore>(
...
+/// Once decrypted, the DEK serves as a symmetric encryption key.
+///
+/// It can also be useful to think of an EDEK as representing a "document access grant" to a user/group.
+#[derive(Debug, Clone, PartialEq)]
+pub struct EncryptedDek {
+    grant_to: WithKey<UserOrGroup>,
+    encrypted_dek_data: recrypt::api::EncryptedValue,
+}
+
+impl From<&EncryptedDek> for EncryptedDekP {
```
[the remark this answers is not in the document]
clintfred: The reason to make take the reference is so that call sites don't have to `clone`. We do this in various places already.

src/internal/document_api/mod.rs
```diff
@@ -435,53 +500,177 @@ pub fn encrypt_document<
...
+                    recrypt,
+                    dek,
+                    encryption_result,
+                    &doc_id,
+                    grants(user_master_pub_key),
+                )?;
+                let proto_edek_vec = encryption_result
+                    .edeks
+                    .iter()
+                    .map(|edek| EncryptedDekP::from(edek))
```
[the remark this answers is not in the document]
clintfred: yes. At one point the type inference was breaking down, but it's working now. /shrug

src/internal/document_api/mod.rs
clintfred: I like that better too. Done.

src/internal/document_api/mod.rs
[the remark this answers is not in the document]
coltfred: I don't think it's worth calling out, but if you insist.

src/internal/document_api/mod.rs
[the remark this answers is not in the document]
coltfred: This function must return a result anyways because of the user or group (my next comment).

src/internal/document_api/mod.rs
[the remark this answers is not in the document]
coltfred: Most of the functions you're calling don't even need a clone, like `.bytes_x_y` so maybe you can get rid of the clone entirely.

src/internal/document_api/mod.rs
coltfred: The structure of the proto is currently the exact same as recrypt. All consumers will have x and y because that's what recrypt provides (and requires). Seems like trying to put them together in the proto has limited gains.

src/internal/document_api/mod.rs
clintfred: You are correct. This function has been refactored a few times (and just was again), and something got messed up along the way. Thanks!

src/internal/document_api/mod.rs
clintfred: I thought the same thing. We keep pulling the x and y along for no reason. If want want to get rid of it, recrypt would be the place, however.

src/document.rs
```diff
@@ -186,6 +186,26 @@ pub trait DocumentOps {
...
+    ///
+    /// The webservice is still needed for looking up public keys and evaluating policies, but no
+    /// document is created and the edeks are not stored. An additional burden is put on the caller
+    /// in that the encrypted data AND the edeks need to be provided for decryption.
+    ///
+    /// # Arguments
+    /// - `document_data` - Bytes of the document to encrypt
+    /// - `encrypt_opts` - Optional document encrypt parameters. Includes
+    ///       `id` - Unique ID to use for the document. Document ID will be stored unencrypted and must be unique per segment.
+    ///       `name` - (Ignored) - Any name provided will be ignored
```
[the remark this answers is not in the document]
clintfred: We don't. Right @skeet70 ? I think this came up in a past discussion and we decided not to include it.

src/document.rs
```diff
@@ -319,6 +339,46 @@ impl DocumentOps for crate::IronOxide {
...
         ))
     }
+
+    fn document_edek_encrypt(
+        &self,
+        data: &[u8],
+        encrypt_opts: &DocumentEncryptOpts,
+    ) -> Result<DocumentDetachedEncryptResult> {
+        let mut rt = Runtime::new().unwrap();
+        let encrypt_opts = encrypt_opts.clone();
```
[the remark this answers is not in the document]
clintfred: No. It was just a big hammer. I have scoped it down to only clone what's necessary.

src/document.rs
```diff
@@ -319,6 +339,46 @@ impl DocumentOps for crate::IronOxide {
...
+            };
+
+        rt.block_on(document_api::edek_encrypt_document(
+            self.device.auth(),
+            &self.recrypt,
+            &self.user_master_pub_key,
+            &self.rng,
+            data,
+            encrypt_opts.id,
+            grant_to_author,
```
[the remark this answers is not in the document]
clintfred: It is part of the Opts for the public api, and the internal API assumes everything is exploded out from the opts, so I'm not sure what you are suggesting here.

src/internal/document_api/mod.rs
clintfred: `encrypt_future` aes encrypted the data. `encrypt_document_core` recrypt encrypts the DEK

src/internal/document_api/mod.rs
clintfred: this was a partially applied function. I've encoded this in a more straightforward way now.

src/internal/document_api/mod.rs
clintfred: Yes. I reworked this slightly so that the clone is not necessary. Nice catch.

src/document.rs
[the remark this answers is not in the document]
BobWall23: It looks like it extracted `grant_to_author` from the `explicit_grants`, rather than from the `encrypt_opts`.

src/internal/document_api/mod.rs
BobWall23: Agree that it shouldn't stay split out in recrypt and get jammed together up higher. But should recrypt keep the internal details internal?

src/internal/document_api/mod.rs
clintfred: Maybe? Much of the work in this branch was to refactor the duplicated encrypt code. The result of these refactorings are `encrypte_document_core` and `resolve_keys_for_grants`. After another glance I don't see any big gains to be had by further code sharing between `edek_encrypt_docuement` and `encrypt_document`. Certainly open to suggestions thought.

src/internal/document_api/mod.rs
BobWall23: `encrypt_document_core` as a name didn't really help me understand that. Seems almost counter-intuitive - `encrypt_document_core` encrypts the stuff that isn't the core of the document.

src/document.rs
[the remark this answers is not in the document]
clintfred: `grant_to_author` is transitively in the `DocumentEncryptOpts`
```rust 
#[derive(Debug, PartialEq, Clone)]
pub struct DocumentEncryptOpts {
    id: Option<DocumentId>,
    name: Option<DocumentName>,
    grants: EitherOrBoth<ExplicitGrant, PolicyGrant>,
}
#[derive(Debug, PartialEq, Clone)]
pub struct ExplicitGrant {
    grant_to_author: bool,
    grants: Vec<UserOrGroup>,
}
```

src/internal/document_api/mod.rs
clintfred: I would argue this is clear from the docs on `generate_new_doc_key`
```rust
/// Generate a DEK and it's associated symmetric key for a new document
pub fn generate_new_doc_key<CR: rand::CryptoRng + rand::RngCore>(
    recrypt: &Recrypt<Sha256, Ed25519, RandomBytes<CR>>,
) -> (Plaintext, DerivedSymmetricKey) {
```

src/internal/document_api/mod.rs
```diff
@@ -250,6 +254,68 @@ impl DocumentMetadataResult {
...
+                }
+            })
+            .collect();
+        result
+    }
+}
+
 /// Result for encrypt operations.
 ///
 /// - `id` - Unique (within the segment) id of the document
```
skeet70: Could you correct the docuemnt misspelling below this line?

src/internal/document_api/mod.rs
clintfred: Ah. It was meant to be the "core" of the encryption not the document. Renamed to `recrypt_document' to see if that makes more sense.

src/internal/document_api/mod.rs
```diff
@@ -435,53 +500,169 @@ pub fn encrypt_document<
...
+
+type UserMasterPublicKey = PublicKey;
+/// Get the public keys for a document grant.
+///
+/// # Arguments
+/// `auth`          - info to make webservice requests
+/// `user_grants`   - list of user ids to which document access should be granted
+/// `group_grants`  - list of groups ids to which document access should be granted
+/// `policy_grant`  - policy to apply for document access
+/// `grant_to_author` - true if document access should be granted to the logged in user, else false
```
giarc3: Is this correct? The type doesn't seem to line up with the last argument in the function (`maybe_user_master_pub_key`)

src/internal/document_api/mod.rs
```diff
@@ -435,53 +500,169 @@ pub fn encrypt_document<
...
-                other_errs,
+            (
+                { [maybe_self_grant, explicit_grants, applied_policy_grants].concat() },
+                [group_errs, user_errs, policy_errs].concat(),
             )
         })
 }
 
+/// Encrypts a document but does not create the document in the IronCore system.
+/// The resultant DocumentDetachedEncryptResult both the EncryptedDeks and the AesEncryptedValue for the caller to deal with.
```
giarc3: Nitpick - this sentence is missing a verb, and it's making it slightly less clear to me. I assume it means "The resultant DocumentDetachedEncryptResult **contains** both the..."?

src/internal/document_api/mod.rs
```diff
@@ -250,6 +254,68 @@ impl DocumentMetadataResult {
...
+        let pb_result: ProtobufResult<EncryptedDeksP> =
+            protobuf::parse_from_bytes(&self.encrypted_deks);
+        let result: Vec<UserOrGroup> = pb_result
+            // safe since the proto-encoded data is encoded in this file using the same definition as this decode
+            .expect("Unable to decode protobuf encoded encrypted_data")
+            .edeks
+            .as_slice()
+            .iter()
+            .map(|edek| {
+                let uog = edek.userOrGroup.clone().unwrap();
```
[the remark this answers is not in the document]
clintfred: this code was reworked

src/internal/mod.rs
```diff
@@ -105,6 +106,10 @@ quick_error! {
         NotGroupAdmin(group_id:group_api::GroupId) {
             display("You're are not an administrator of group '{}'", group_id.0)
         }
+        /// Protobuf encode/decode error
```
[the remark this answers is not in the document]
clintfred: The comment does show up on docs.rs, but you might be right. I added another proto error variant and renamed this one, so maybe it has more value now?

src/internal/rest.rs
```diff
@@ -511,49 +511,51 @@ pub mod json {
...
-                        public_signing_key: public_signing_key.bytes().to_vec(),
-                    },
-                }),
-                _ => Err(IronOxideErr::InvalidRecryptEncryptedValue(
-                    "Expected a TransformedValue but got an EncryptedOnceValue".to_string(),
-                )),
-            }
-        }
-    }
+    // TODO Can we delete this?
```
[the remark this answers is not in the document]
clintfred: moved to the `test` mod

src/internal/document_api/mod.rs
clintfred: Thanks. Fixed.

src/internal/document_api/mod.rs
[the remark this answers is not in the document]
clintfred: Function now returns a Result. I think I found a decent way to unpack the users/groups from the proto

src/internal/document_api/mod.rs
```diff
@@ -250,6 +254,68 @@ impl DocumentMetadataResult {
...
+                    UserOrGroup::User {
+                        id: UserId::unsafe_from_string(uog.get_userId().to_string()),
+                    }
+                } else if uog.has_groupId() {
+                    UserOrGroup::Group {
+                        id: GroupId::unsafe_from_string(uog.get_groupId().to_string()),
+                    }
+                } else {
+                    // a userOrGroup must be either a user or a group
+                    unreachable!()
```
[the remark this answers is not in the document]
clintfred: figured out a way to use pattern matching

src/internal/document_api/mod.rs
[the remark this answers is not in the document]
clintfred: I now only clone the grant_to, but not the edeks

src/internal/mod.rs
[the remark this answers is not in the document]
skeet70: I mean, the error name is now even more descriptive now, I think it has even less value :D

src/internal/document_api/mod.rs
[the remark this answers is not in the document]
clintfred: This section of code was refactored

src/internal/document_api/mod.rs
clintfred: test was added

proto/transform.proto
```diff
@@ -0,0 +1,37 @@
...
+    oneof UserOrGroupId {
+        string userId = 1;
+        string groupId = 2;
+    }
+
+    PublicKey masterPublicKey = 3;
+}
+
+message EncryptedDekData {
+    bytes encryptedMessage = 1; // encrypted_dek_bytes
```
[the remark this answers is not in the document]
clintfred: I would like to rename `encryptedMessage` to `encryptedDekBytes`. In recrypt, it's an encryptedMessage makes sense as it's a more generic usage. Here we know what these bytes are. 
@coltfred thoughts?

src/internal/mod.rs
```diff
@@ -171,7 +176,7 @@ pub struct RequestAuth {
     ///The users given id, which uniquely identifies them inside the segment.
     account_id: UserId,
     ///The segment_id for the above user.
-    segment_id: usize,
+    segment_id: usize, //TODO this should be i32 to match the webservice
```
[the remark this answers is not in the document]
clintfred: I rabbit-holed trying to do this a couple of times and both times backed out about 10 minutes in. I'll write it up as a separate ticket.

src/internal/document_api/mod.rs
clintfred: I am very convinced that splitting the public key into x and y for public consumption was a mistake. We may have passed some point of no return as this design choice reverberates through many SDKs, REST endpoints, and serialization choices. We _could_ draw the line at edeks, combine these into on 64 byte blob and write some issues to clean that up in other places. IronOxide already supports serializing as a single array. Recrypt _could_ support in/out of a combined array and we could deprecate the other one.

We have to start somewhere... and it will never be easier than it is today to clean this up. @coltfred would you support making this change in the proto?

src/internal/document_api/mod.rs
```diff
@@ -435,53 +508,163 @@ pub fn encrypt_document<
...
-            } else {
-                [&users_with_key[..], &groups_with_key[..]].concat()
-            };
+            let explicit_grants = [users_with_key, groups_with_key].concat();
             let (policy_errs, applied_policy_grants) = match maybe_policy_res {
                 None => (vec![], vec![]),
                 Some(res) => process_policy(&res),
             };
+            let maybe_self_grant = {
+                if let Some(user_master_pub_key) = maybe_user_master_pub_key {
```
coltfred: To me this makes more sense to do this with a `map_or`, but if that doesn't compute for you I won't be hurt.

src/internal/document_api/mod.rs
```diff
@@ -1054,4 +1343,105 @@ mod tests {
...
+                    .get_publicSigningKey()
+                    .to_vec()
+            );
+
+            assert_eq!(
+                signature.bytes().to_vec(),
+                proto_edek.get_encryptedDekData().get_signature().to_vec()
+            );
+        } else {
+            panic!("Should be EncryptedOnceValue");
```
coltfred: Shouldn't this just be a test failure not a panic?

src/lib.rs
```diff
@@ -27,6 +27,9 @@
 //! groups. Groups can be created, updated, and deleted along with management of a groups administrators and members.
 //!
 
+// required by quick_error or IronOxideErr
+#![recursion_limit = "128"]
```
coltfred: :confused: Why?

tests/document_ops.rs
```diff
@@ -211,32 +212,209 @@ fn doc_create_with_policy_grants() -> Result<(), IronOxideErr> {
 }
 
 #[test]
-fn doc_create_with_explicit_grants() {
-    let sdk = init_sdk();
+fn doc_edek_encrypt_with_policy_grants() -> Result<(), IronOxideErr> {
```
coltfred: This test name should reflect the name changes.

tests/document_ops.rs
```diff
@@ -211,32 +212,209 @@ fn doc_create_with_policy_grants() -> Result<(), IronOxideErr> {
...
+    // policy assumed for this test
+    /*
```
coltfred: I don't feel like we need another test on policy. I know it's tempting to just duplicate the other tests for this, but it causes any changes to make changing the tests onerous. Is there a way we can simplify this test?

src/internal/document_api/mod.rs
clintfred: I can see that, but I think I'll just leave it as is.

src/internal/document_api/mod.rs
clintfred: `assert!()` is just a `panic!` with a boolean check, so I think it's okay. Maybe there is just a fail macro? I'll check.

src/lib.rs
clintfred: quick_error macro must process the errors recursively? All I know is that I added an error and I broke the standard limit of 64/

coltfred: huh, ok.

tests/document_ops.rs
clintfred: After review, I agree. Deleted!

src/internal/document_api/mod.rs
clintfred: We are punting this for now.

src/internal/document_api/mod.rs
```diff
@@ -250,10 +254,80 @@ impl DocumentMetadataResult {
...
+        &self.encrypted_deks
+    }
+    pub fn access_errs(&self) -> &[DocAccessEditErr] {
+        &self.access_errs
+    }
+
+    /// Users and Groups that have access to the encrypted document.
+    ///
+    /// Implementation is somewhat expensive as it is decoding the proto encoded `encrypted_deks`.
+    pub fn grants(&self) -> Result<Vec<UserOrGroup>, IronOxideErr> {
```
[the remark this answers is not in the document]
clintfred: I ended up just caching the grants on the result creation -- which was what I was originally trying to avoid. Deserializing proto just makes the interface too ugly...

PR #32, opened 2019-08-15
[#22] Unmanaged Decryption

clintfred:
see #22 

## Changelog 
`DocumentAdvancedOps::document_decrypt_unmanaged` function added for advanced use cases. This decrypt operation is the inverse of `DocumentAdvancedOps::document_encrypt_unmanaged`

### DIFF
--- a/src/document/advanced.rs
+++ b/src/document/advanced.rs
@@ -28,6 +28,7 @@ pub trait DocumentAdvancedOps {
         encrypt_opts: &DocumentEncryptOpts,
     ) -> Result<DocumentEncryptUnmanagedResult>;
 
+    /// (Advanced) TODO
     fn document_decrypt_unmanaged(
         &self,
         encrypted_data: &[u8],
@@ -80,10 +81,6 @@ impl DocumentAdvancedOps for crate::IronOxide {
         encrypted_data: &[u8],
         encrypted_deks: &[u8],
     ) -> Result<DocumentDecryptUnmanagedResult> {
-        let deks: crate::proto::transform::EncryptedDeks =
-            protobuf::parse_from_bytes(&encrypted_deks)?;
-        dbg!(&deks);
-
         let mut rt = Runtime::new().unwrap();
 
         rt.block_on(internal::document_api::decrypt_document_unmanaged(
--- a/src/document/advanced.rs
+++ b/src/document/advanced.rs
@@ -1,6 +1,8 @@
 use crate::document::{partition_user_or_group, DocumentEncryptOpts};
 use crate::internal;
-use crate::internal::document_api::DocumentEncryptUnmanagedResult;
+pub use crate::internal::document_api::{
+    DocumentDecryptUnmanagedResult, DocumentEncryptUnmanagedResult,
+};
 use crate::Result;
 use itertools::EitherOrBoth;
 use tokio::runtime::current_thread::Runtime;
@@ -25,6 +27,12 @@ pub trait DocumentAdvancedOps {
         data: &[u8],
         encrypt_opts: &DocumentEncryptOpts,
     ) -> Result<DocumentEncryptUnmanagedResult>;
+
+    fn document_decrypt_unmanaged(
+        &self,
+        encrypted_data: &[u8],
+        encrypted_deks: &[u8],
+    ) -> Result<DocumentDecryptUnmanagedResult>;
 }
 
 impl DocumentAdvancedOps for crate::IronOxide {
@@ -64,6 +72,26 @@ impl DocumentAdvancedOps for crate::IronOxide {
             &explicit_users,
             &explicit_groups,
             policy_grants,
+        ))
+    }
+
+    fn document_decrypt_unmanaged(
+        &self,
+        encrypted_data: &[u8],
+        encrypted_deks: &[u8],
+    ) -> Result<DocumentDecryptUnmanagedResult> {
+        let deks: crate::proto::transform::EncryptedDeks =
+            protobuf::parse_from_bytes(&encrypted_deks)?;
+        dbg!(&deks);
+
+        let mut rt = Runtime::new().unwrap();
+
+        rt.block_on(internal::document_api::decrypt_document_unmanaged(
+            self.device.auth(),
+            &self.recrypt,
+            self.device().private_device_key(),
+            encrypted_data,
+            encrypted_deks,
         ))
     }
 }
--- a/src/document/advanced.rs
+++ b/src/document/advanced.rs
@@ -28,7 +28,15 @@ pub trait DocumentAdvancedOps {
         encrypt_opts: &DocumentEncryptOpts,
     ) -> Result<DocumentEncryptUnmanagedResult>;
 
-    /// (Advanced) TODO
+    /// (Advanced) Decrypt a document not managed by the ironcore service. Both the encrypted
+    /// data and the encrypted deks need to be provided.
+    ///
+    /// The webservice is still needed to transform a chosen encrypted dek so it can be decrypted
+    /// by the caller's private key.
+    ///
+    /// # Arguments
+    /// - `encrypted_data` - Encrypted document
+    /// - `encrypted_deks` - Associated encrypted DEKs for the `encrypted_data`
     fn document_decrypt_unmanaged(
         &self,
         encrypted_data: &[u8],
--- a/src/internal/document_api/mod.rs
+++ b/src/internal/document_api/mod.rs
@@ -95,9 +95,9 @@ struct DocumentHeader {
     pub segment_id: usize,
 }
 
-// Take an encrypted document and extract out the header metadata. Return that metadata as well as the AESEncryptedValue
-// that contains the AES IV and encrypted content. Will fail if the provided document doesn't contain the latest version
-// which contains the header bytes.
+/// Take an encrypted document and extract out the header metadata. Return that metadata as well as the AESEncryptedValue
+/// that contains the AES IV and encrypted content. Will fail if the provided document doesn't contain the latest version
+/// which contains the header bytes.
 fn parse_document_parts(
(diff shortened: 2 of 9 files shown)


src/internal/document_api/mod.rs
```diff
@@ -795,18 +789,69 @@ impl TryFrom<&EncryptedDek> for EncryptedDekP {
...
+        ]
+        .concat()
+    }
+
+    /// vector of EDEKs
+    fn edek_vec(&self) -> Vec<EncryptedDek> {
+        self.value.edeks.clone()
+    }
+
+    fn edek_bytes(&self) -> Result<Vec<u8>, IronOxideErr> {
```
clintfred: function doc here

src/internal/document_api/mod.rs
```diff
@@ -795,18 +789,69 @@ impl TryFrom<&EncryptedDek> for EncryptedDekP {
...
 #[derive(Debug, Clone)]
-pub struct EncryptionResult {
+struct RecryptionResult {
     edeks: Vec<EncryptedDek>,
     encrypted_data: AesEncryptedValue,
     encryption_errs: Vec<DocAccessEditErr>,
 }
 
+impl RecryptionResult {
+    fn into_edoc_unmanaged(self, header: DocumentHeader) -> EncryptedDoc {
```
clintfred: better name would be into_edoc

src/internal/document_api/mod.rs
```diff
@@ -915,6 +944,71 @@ pub fn decrypt_document<'a, CR: rand::CryptoRng + rand::RngCore>(
...
+                        id: document_id,
+                        access_via: user_or_group,
+                        decrypted_data: DecryptedData(decrypted_doc.to_vec()),
+                    })
+            },
+        )
+}
+
+struct DecryptedData(Vec<u8>);
+pub struct DocumentDecryptUnmanagedResult {
```
clintfred: documentation block needed here

tests/document_ops.rs
```diff
@@ -606,7 +621,7 @@ fn doc_grant_access() {
             },
         ],
     );
-    dbg!(&grant_result);
+    //    dbg!(&grant_result);
```
clintfred: remove

src/internal/document_api/mod.rs
```diff
@@ -915,6 +944,71 @@ pub fn decrypt_document<'a, CR: rand::CryptoRng + rand::RngCore>(
...
+/// encrypted DEKs. Most use cases would want `decrypt_document` instead.
+pub fn decrypt_document_unmanaged<'a, CR: rand::CryptoRng + rand::RngCore>(
+    auth: &'a RequestAuth,
+    recrypt: &'a Recrypt<Sha256, Ed25519, RandomBytes<CR>>,
+    device_private_key: &'a PrivateKey,
+    encrypted_doc: &'a [u8],
+    encrypted_deks: &'a [u8],
+) -> impl Future<Item = DocumentDecryptUnmanagedResult, Error = IronOxideErr> + 'a {
+    // attempt to parse the proto as fail-fast validation. If it fails decrypt will fail
+    protobuf::parse_from_bytes::<EncryptedDeksP>(encrypted_deks)
```
clintfred: Since we are going to pay the cost of proto de-serialization, we could also check that document_id and segment_id of the edoc and edeks match prior to making a network call as additional validation.

clintfred: I did this.

src/internal/document_api/mod.rs
```diff
@@ -1469,11 +1598,53 @@ mod tests {
             aes_value,
             &doc_id,
             with_keys,
-        )?;
+        )?
+        .into_edoc(DocumentHeader::new(doc_id.clone(), seg_id));
```
BobWall23: Isn't `new` going to clone the input again to construct the object?

src/internal/document_api/mod.rs
```diff
@@ -1469,11 +1598,53 @@ mod tests {
...
+
+        let edoc_bytes = encryption_result.edoc_bytes();
+        let (parsed_header, _) = parse_document_parts(&edoc_bytes)?;
+        assert_eq!(&encryption_result.header, &parsed_header);
+
+        Ok(())
+    }
+
+    #[test]
+    pub fn unmanaged_edec_compare_grants() -> Result<(), IronOxideErr> {
```
BobWall23: s/edec/edoc/ ?

src/internal/document_api/mod.rs
```diff
@@ -86,18 +86,45 @@ impl TryFrom<&str> for DocumentName {
...
+
+impl DocumentHeader {
+    fn new(document_id: DocumentId, segment_id: usize) -> DocumentHeader {
+        DocumentHeader {
+            document_id,
+            segment_id,
+        }
+    }
+    /// Generate a documents header given its ID and internal segment ID that is is associated with. Generates
+    /// a Vec<u8> which includes the document version, header size, and header JSON as bytes.
```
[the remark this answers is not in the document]
clintfred: I just moved this code from elsewhere in this file. This is for version 2 documents (pre-EDOC). I don't think any SDKs currently support EDOC (v3).

src/internal/document_api/mod.rs
```diff
@@ -86,18 +86,45 @@ impl TryFrom<&str> for DocumentName {
...
+        DocHeaderPacked(header)
+    }
 }
 
-// Take an encrypted document and extract out the header metadata. Return that metadata as well as the AESEncryptedValue
-// that contains the AES IV and encrypted content. Will fail if the provided document doesn't contain the latest version
-// which contains the header bytes.
+/// Take an encrypted document and extract out the header metadata. Return that metadata as well as the AESEncryptedValue
+/// that contains the AES IV and encrypted content. Will fail if the provided document doesn't contain the latest version
+/// which contains the header bytes.
```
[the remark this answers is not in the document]
clintfred: IronOxide only reads version 2 documents. I guess we should write a ticket for EDOC support.

src/internal/document_api/mod.rs
clintfred: In Rust this is implementation specific. This "new" looks like:
```rust 
fn new(document_id: DocumentId, segment_id: usize) -> DocumentHeader {
```
This means that the new function takes ownership of both arguments so if you want to use `doc_id` again in this function we have to give `new` its own copy.

src/internal/document_api/mod.rs
```diff
@@ -915,6 +946,100 @@ pub fn decrypt_document<'a, CR: rand::CryptoRng + rand::RngCore>(
         })
 }
 
+/// Decrypt the unmanaged document. The caller must provide both the encrypted data as well as the
+/// encrypted DEKs. Most use cases would want `decrypt_document` instead.
+pub fn decrypt_document_unmanaged<'a, CR: rand::CryptoRng + rand::RngCore>(
```
giarc3: Is there a reason that these methods are named with like `decrypt_document_unmanaged`, but other repos (and the Result classes) are named like `document_decrypt_unmanaged`

clintfred: The main reason is consistency with the already existing `decrypt_document`. The encrypt functions also have this "backwards" naming convention.

I have added this to https://github.com/IronCoreLabs/ironoxide/issues/33

src/document/advanced.rs
```diff
@@ -25,6 +27,21 @@ pub trait DocumentAdvancedOps {
...
+    ///
+    /// The webservice is still needed to transform a chosen encrypted dek so it can be decrypted
+    /// by the caller's private key.
+    ///
+    /// # Arguments
+    /// - `encrypted_data` - Encrypted document
+    /// - `encrypted_deks` - Associated encrypted DEKs for the `encrypted_data`
+    fn document_decrypt_unmanaged(
+        &self,
+        encrypted_data: &[u8],
```
coltfred: We have these in the opposite order in IronWeb. Can we reorder them?

src/internal/document_api/mod.rs
```diff
@@ -795,18 +809,71 @@ impl TryFrom<&EncryptedDek> for EncryptedDekP {
...
+        self.value.edeks.clone()
+    }
+
+    /// binary blob for associated edeks, or error if encoding the edeks failed
+    fn edek_bytes(&self) -> Result<Vec<u8>, IronOxideErr> {
+        let proto_edek_vec_results: Result<Vec<_>, _> = self
+            .value
+            .edeks
+            .iter()
+            .map(|edek| edek.try_into())
```
coltfred: I'm surprised going _to_ proto can fail...

clintfred: It can only fail because an internal `EncryptedDek` could be holding a `TransformedValue` because of the `EncryptedValue` recrypt enum. This would be a bug in IronOxide if this ever happened. This is only happening on umanaged encrypt. Previously it was being masked there as that has to return a `Result` for other errors as well.

We can relax this if enum variants ever become "real" types.

src/document/advanced.rs
clintfred: Slight push-back... If we think of unmanaged decrypt as a variant of normal decrypt it makes more sense to me to add the new parameter onto the end of the function call.

coltfred: Changed IronWeb.

src/internal/document_api/mod.rs
coltfred: Ugh. That's too bad.

src/internal/document_api/mod.rs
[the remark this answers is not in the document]
clintfred: Detached EDEKs and EDOCv3 are decoupled.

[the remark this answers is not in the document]
clintfred: written up as https://github.com/IronCoreLabs/ironoxide/issues/34

Issue #38, opened 2019-09-11
Create user with needsRotation set


clintfred:
As an application developer
I want to be able to create a user's cryptographic identity ahead of them logging in
so that I can add them to groups and/or share data with them.

AC:
* [ ] - user_create takes an options object that allows a user to created with a flag set marking them as needing rotation
* [ ] - `needsRotation` is returned as part of the value returned to the caller

PR #43, opened 2019-09-16
[#38] Support user create with needs_rotation set

clintfred:
see #38 

## Changelog
* Users can now be created with a `needs_rotation` flag set


src/user.rs
```diff
@@ -99,22 +130,27 @@ pub trait UserOps {
...
             password.try_into()?,
+            user_create_opts.needs_rotation,
             OUR_REQUEST,
         ))
     }
 
+    fn user_list_devices(&self) -> Result<UserDeviceListResult> {
+        let mut rt = Runtime::new().unwrap();
+        rt.block_on(user_api::device_list(self.device.auth()))
+    }
```
ernieturner: Did we just forget to expose this function or something?

src/user.rs
```diff
@@ -29,18 +29,49 @@ impl Default for DeviceCreateOpts {
...
+    /// - `user_create_opts` - see [`UserCreateOpts`](struct.UserCreateOpts.html)
     /// # Returns
     /// Newly generated `UserCreateKeyPair` or Err. For most use cases this key pair can
     /// be discarded as IronCore escrows your user's keys. The escrowed keys are unlocked
     /// by the provided password.
-    fn user_create(jwt: &str, password: &str) -> Result<UserCreateKeyPair>;
+    fn user_create(
+        jwt: &str,
+        password: &str,
+        user_create_opts: &UserCreateOpts,
```
ernieturner: I assume it's a purposeful choice to break this API and not make this `Option<UserCreateOpts>` instead? 

Related, should you change the version to `0.11.0` in the Cargo.toml as part of this PR?

src/user.rs
```diff
@@ -29,18 +29,49 @@ impl Default for DeviceCreateOpts {
...
+impl UserCreateOpts {
+    /// Constructor. Also see [`default`](#method.default).
+    ///
+    /// # Arguments
+    /// - `needs_rotation` - Set to true if the private key for this user should be rotated when the user "takes control" of their keys.
+    ///
+    /// The main use case for this is a workflow that requires that users and groups to be generated
+    /// prior to the user logging in for the first time. In this situation, a user's cryptographic identity
+    /// can be generated by a third party, like a server process, and then the user can take control of their
+    /// keys by rotating the private key using [`rotate_private_key`](TODO LINK).
```
coltfred: Maybe we should make an issue for this and link to that here.

src/user.rs
coltfred: No, it just got moved.

clintfred: Sorry for the confusion. We were getting a linting warning about our method order not matching the trait order. I fixed it in this commit: https://github.com/IronCoreLabs/ironoxide/pull/43/commits/c68eb406b6054a2322c1d456339b5fe4602f5aec

Churn!

src/user.rs
clintfred: Yes, intentional, but good point on the version bump.

src/user.rs
clintfred: added a link to https://github.com/IronCoreLabs/ironoxide/issues/44

PR #56, opened 2019-10-18
Add needs_rotation to group responses

giarc3:
- Added needs_rotation as an `Option<bool>` to:
    - `GroupMetaResult`
    - `GroupGetResult`
    - `GroupBasicApiResponse`
    - `GroupGetApiResponse`
- `GroupListResult` is simply a `Vec<GroupMetaResult>`, so it will get these changes.
- Added a test for `group_get_metadata()` that tests a call by an admin, member, and non-member.

`needs_rotation` will be a `None` for a non-admin, and a `Some` for an admin.

### DIFF
--- a/src/internal/group_api/mod.rs
+++ b/src/internal/group_api/mod.rs
@@ -177,7 +177,8 @@ impl GroupGetResult {
     pub fn member_list(&self) -> Option<&Vec<UserId>> {
         self.member_list.as_ref()
     }
-    /// None if the calling user is not a group admin
+    /// `Some(boolean)` indicating if the group needs rotation if the calling user is a group admin.
+    /// `None` if the calling user is not a group admin.
     pub fn needs_rotation(&self) -> Option<bool> {
         self.needs_rotation
     }
--- a/src/internal/group_api/mod.rs
+++ b/src/internal/group_api/mod.rs
@@ -124,6 +124,8 @@ impl GroupMetaResult {
     pub fn group_master_public_key(&self) -> &PublicKey {
         &self.group_master_public_key
     }
+    /// `Some(boolean)` indicating if the group needs rotation if the calling user is a group admin.
+    /// `None` if the calling user is not a group admin.
     pub fn needs_rotation(&self) -> Option<bool> {
         self.needs_rotation
     }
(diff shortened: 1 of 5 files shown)


tests/group_ops.rs
```diff
@@ -18,16 +18,36 @@ fn group_create_no_member() {
...
+fn group_get_metadata() {
+    let admin_sdk = init_sdk();
+    let member_sdk = init_sdk();
+    let nonmember_sdk = init_sdk();
+
+    let member_id = member_sdk.device().account_id().clone();
+    let group = admin_sdk.group_create(&Default::default());
+    let group_id = &group.unwrap().id().clone();
+
+    let _ = admin_sdk.group_add_members(&group_id, &[member_id]);
```
coltfred: I don't think you need the let _ =

src/internal/group_api/mod.rs
```diff
@@ -172,6 +177,10 @@ impl GroupGetResult {
     pub fn member_list(&self) -> Option<&Vec<UserId>> {
         self.member_list.as_ref()
     }
+    /// None if the calling user is not a group admin
```
clintfred: Should explain the other two states: Some(true), Some(false)

tests/group_ops.rs
clintfred: I think you will get a compiler warning about an unused Result if you ignore the result...

clintfred: I see now that Craig used `?` instead. That's a good option. :+1:

src/internal/group_api/mod.rs
giarc3: I was/am having difficulty wording it succinctly. I made an attempt at a comment, but I am open to suggested changes.

src/internal/group_api/mod.rs
```diff
@@ -123,6 +124,11 @@ impl GroupMetaResult {
     pub fn group_master_public_key(&self) -> &PublicKey {
         &self.group_master_public_key
     }
+    /// `Some(boolean)` indicating if the group needs rotation if the calling user is a group admin.
+    /// `None` if the calling user is not a group admin.
```
clintfred: This is reasonable. My attempt that this documentation block would be:
```
/// `None` - calling user is not a group admin and cannot see if the group needs rotation or not
/// `Some(true)` - calling user is an admin and the group needs rotation
/// `Some(false)` - calling user is an admin and the group does not need rotation
```

Issue #44, opened 2019-09-18
Support user private key rotation


clintfred:
As a user
I want to be able to rotate my private key (without changing my public key)
so that I can be sure that anyone who had access to my private key previously, can't decrypt data

AC:
* [x] new init function exists of the form `pub fn initialize_check_rotation(device_context: &DeviceContext) -> Result<InitAndRotationCheck>`
* [x] new function in the user interface `fn rotate_curr_users_priv_key(&self, password: &str) -> Result<UserPrivateKeyRotationResult>`
* [x] calling `fn rotate_curr_users_priv_key` should result in the user's private key being changed, and old version of the private key should be useless. Call the `PUT users/{userId}/keys/{userKeyId}` in the ironcore service. 
* [x] after the private key is rotated, existing device transform keys should continue to work
* [x] after private key rotation, the `needsRotation` flag should be unset on subsequent `initialilze` calls (and on other calls that can get `needsRotation`

related to https://github.com/IronCoreLabs/ironweb/issues/38

clintfred:
When picking the augmenting factor, the caller should make sure the current private key minus the augmentation factor is not 0.

clintfred:
Also need to make sure the `AugmentingFactor` itself is not zero. This could potentially happen if a sha-256 hash can ever be all zero.

Issue #45, opened 2019-09-19
Support group private key rotation


clintfred:
As a group admin
I want to be able to rotate my private key (without changing my public key)
so that I can be sure that anyone who had access to the old private key can't administer the group after rotation.

AC: 
* [ ] if the user elects to check for group private key rotation on initialization, groups that need rotation will be returned to them
* [ ] new function exists in the group interface `fn rotate_group_priv_key(id: GroupId) -> Result<GroupPrivateKeyRotationResult>`
* [ ] calls the `PUT groups/{groupId}/keys/{groupKeyId}` in the ironcore service
* [ ] after the group's private key is rotated prior group private keys should not be able to be used for group administration operations
* [ ] group's currentKeyid added to GroupBasicApiResponse, GroupApiResponse (currentKeyId should not be part of the public API)

PR #59, opened 2019-10-22
44 user private key rotate

clintfred:
See #44 

## TODO
* resolve issue with generating devices after user private key rotation

## Changelog
* Adds method `UserOps::user_rotate_private_key`
* Adds a new initialization option: `ironoxide::initialize_check_rotation` to enable users to know if any of their private keys need rotation.

### DIFF
--- a/src/internal/mod.rs
+++ b/src/internal/mod.rs
@@ -487,7 +487,7 @@ impl<'de> Deserialize<'de> for PrivateKey {
 
 impl PrivateKey {
     /// Augment this private key with another, producing a new PrivateKey
-    fn augment(&self, augmenting_key: AugmentationFactor) -> Result<PrivateKey, IronOxideErr> {
+    fn augment(&self, augmenting_key: &AugmentationFactor) -> Result<PrivateKey, IronOxideErr> {
         use recrypt::Revealed;
         let zero: RecryptPrivateKey = RecryptPrivateKey::new([0u8; 32]);
         if Revealed((augmenting_key.clone().0).0) == Revealed(zero.clone()) {
@@ -495,7 +495,7 @@ impl PrivateKey {
                 "Augmenting key cannot be zero".into(),
             ))
         } else {
-            let augmented_key = self.0.clone() - augmenting_key.into();
+            let augmented_key = self.0.clone() - augmenting_key.clone().into();
             // TODO might be nice if Revealed could hold a reference
             if Revealed(augmented_key.clone()) == Revealed(zero) {
                 Err(IronOxideErr::UserPrivateKeyRotationError(
@@ -515,6 +515,10 @@ impl AugmentationFactor {
     pub fn generate_new<CO: CryptoOps>(recrypt: &CO) -> AugmentationFactor {
         let pt = recrypt.gen_plaintext();
         AugmentationFactor(recrypt.derive_private_key(&pt).into())
+    }
+
+    pub fn as_bytes(&self) -> &[u8; 32] {
+        self.0.as_bytes()
     }
 }
 
@@ -948,7 +952,7 @@ pub(crate) mod test {
     fn private_key_augment_with_self_is_none() {
         let privk = gen_priv_key();
 
-        let result = privk.augment(AugmentationFactor(privk.clone()));
+        let result = privk.augment(&AugmentationFactor(privk.clone()));
         assert_that!(&result, is_variant!(Err));
         assert_that!(
             &result.unwrap_err(),
@@ -964,7 +968,7 @@ pub(crate) mod test {
 
         let p3 = (p1.clone().0 - p2.clone().0).into();
 
-        let aug_p = p1.augment(AugmentationFactor(p2)).unwrap();
+        let aug_p = p1.augment(&AugmentationFactor(p2)).unwrap();
         assert_eq!(Revealed(aug_p.0), Revealed(p3))
     }
 
@@ -972,7 +976,7 @@ pub(crate) mod test {
     fn private_key_augmentation_aug_key_of_zero() {
         let priv_key_orig = gen_priv_key();
         let zero_aug_factor = AugmentationFactor(PrivateKey(RecryptPrivateKey::new([0u8; 32])));
-        let new_priv_key = priv_key_orig.augment(zero_aug_factor);
+        let new_priv_key = priv_key_orig.augment(&zero_aug_factor);
         assert_that!(&new_priv_key, is_variant!(Err));
         assert_that!(
             &new_priv_key.unwrap_err(),
--- a/src/internal/mod.rs
+++ b/src/internal/mod.rs
@@ -44,6 +44,7 @@ pub enum RequestErrorCode {
     UserDeviceDelete,
     UserDeviceList,
     UserKeyList,
+    UserKeyUpdate,
     GroupCreate,
     GroupDelete,
     GroupList,
--- a/src/internal/mod.rs
+++ b/src/internal/mod.rs
@@ -123,6 +123,9 @@ quick_error! {
             Encrypted Document is DocumentId({}) and SegmentId({})",
             edek_doc_id, edek_segment_id, edoc_doc_id, edoc_segment_id)
         }
+        UserPrivateKeyRotationError(msg: String) {
+            display("User private key rotation failed with '{}'", msg)
+        }
     }
 }
 
@@ -400,6 +403,30 @@ impl<'de> Deserialize<'de> for PrivateKey {
         let s = String::deserialize(deserializer)?;
         let keys_bytes = base64::decode(&s).map_err(|e| Error::custom(e.to_string()))?;
         PrivateKey::try_from(&keys_bytes[..]).map_err(|e| Error::custom(e.to_string()))
+    }
+}
+
+impl PrivateKey {
+    /// Augment this private key with another, producing An AugmentationFactor
+    fn augment(&self, augmenting_key: PrivateKey) -> Option<AugmentationFactor> {
+        use recrypt::Revealed;
+        let zero: RecryptPrivateKey = RecryptPrivateKey::new([0u8; 32]);
+        let augmented_key = self.0.clone() - augmenting_key.0;
+        // TODO might be nice if Revealed could hold a reference
+        if Revealed(augmented_key.clone()) == Revealed(zero) {
+            None
+        } else {
+            Some(AugmentationFactor(augmented_key.into()))
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct AugmentationFactor(PrivateKey);
+
+impl From<AugmentationFactor> for PrivateKey {
+    fn from(aug: AugmentationFactor) -> Self {
+        aug.0
     }
 }
 
@@ -781,5 +808,31 @@ pub(crate) mod test {
             (&proto_pubk.get_x().to_vec(), &proto_pubk.get_y().to_vec())
         );
         Ok(())
+    }
+
+    pub fn gen_priv_key() -> PrivateKey {
+        let recr = recrypt::api::Recrypt::new();
+        let (re_privk, _) = recr.generate_key_pair().unwrap();
+        re_privk.into()
+    }
+
+    #[test]
+    fn private_key_augment_with_self_is_none() {
+        let privk = gen_priv_key();
+
+        let result = privk.augment(privk.clone());
+        assert!(&result.is_none())
+    }
+
+    #[test]
+    fn private_key_augmentation_is_subtraction() {
+        use recrypt::Revealed;
+        let p1 = gen_priv_key();
+        let p2 = gen_priv_key();
+
+        let p3 = (p1.clone().0 - p2.clone().0).into();
+
+        let aug_p = p1.augment(p2);
+        assert_eq!(Revealed((aug_p.unwrap().0).0), Revealed(p3))
     }
 }
--- a/src/internal/mod.rs
+++ b/src/internal/mod.rs
@@ -496,6 +496,7 @@ impl PrivateKey {
                 "Augmenting key cannot be zero".into(),
             ))
         } else {
+            //TODO is this the right direction?
             let augmented_key = self.0.clone() - augmenting_key.clone().into();
             // TODO might be nice if Revealed could hold a reference
             if Revealed(augmented_key.clone()) == Revealed(zero) {
--- a/src/internal/mod.rs
+++ b/src/internal/mod.rs
@@ -496,7 +496,7 @@ impl PrivateKey {
                 "Augmenting key cannot be zero".into(),
             ))
         } else {
-            //TODO is this the right direction?
+            // this subtractions needs to be the additive inverse of what the service is doing
             let augmented_key = self.0.clone() - augmenting_key.clone().into();
             // TODO might be nice if Revealed could hold a reference
             if Revealed(augmented_key.clone()) == Revealed(zero) {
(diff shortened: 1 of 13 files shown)


giarc3:
I pushed a commit that put `UserUpdatePrivateKeyResult` and `EncryptedPrivateKey` into the `pub use crate` section of `user.rs` so that I could access them in ironoxide-java. Let me know if there's a reason we shouldn't do this and I'll change it back.

src/internal/mod.rs
```diff
@@ -477,6 +485,52 @@ impl<'de> Deserialize<'de> for PrivateKey {
     }
 }
 
+impl PrivateKey {
+    /// Augment this private key with another, producing a new PrivateKey
+    fn augment(&self, augmenting_key: &AugmentationFactor) -> Result<PrivateKey, IronOxideErr> {
+        use recrypt::Revealed;
+        let zero: RecryptPrivateKey = RecryptPrivateKey::new([0u8; 32]);
+        if Revealed((augmenting_key.clone().0).0) == Revealed(zero.clone()) {
```
BobWall23: Why does `Revealed` take ownership of its input, instead of borrowing? Seems like the clones shouldn't be necessary.

src/internal/mod.rs
```diff
@@ -477,6 +485,52 @@ impl<'de> Deserialize<'de> for PrivateKey {
...
+            Err(IronOxideErr::UserPrivateKeyRotationError(
+                "Augmenting key cannot be zero".into(),
+            ))
+        } else {
+            // this subtraction needs to be the additive inverse of what the service is doing
+            let augmented_key = self.0.clone() - augmenting_key.clone().into();
```
BobWall23: Could have compared `self` and `augmenting_key` and errored if they were the same, before doing the subtraction. But with the coming change to do this in Fr instead of Fp, you would have to do that comparison after converting self and augmenting_key to Fr elements.

src/internal/mod.rs
```diff
@@ -477,6 +485,52 @@ impl<'de> Deserialize<'de> for PrivateKey {
...
+}
+
+/// Private key used to augment another PrivateKey
+#[derive(Clone, Debug)]
+pub(crate) struct AugmentationFactor(PrivateKey);
+
+impl AugmentationFactor {
+    /// Use recrypt to generate a new AugmentationFactor
+    pub fn generate_new<CO: CryptoOps>(recrypt: &CO) -> AugmentationFactor {
+        let pt = recrypt.gen_plaintext();
```
BobWall23: could just use `gen_priv_key()` - that would be less work.

src/internal/user_api/mod.rs
```diff
@@ -236,6 +237,99 @@ pub fn user_create<CR: rand::CryptoRng + rand::RngCore>(
...
+            Ok({
+                let password_string = password.0;
+                let encrypt_priv_key = curr_user.user_private_key;
+                let priv_key: PrivateKey = aes::decrypt_user_master_key(
+                    &password_string,
+                    &aes::EncryptedMasterKey::new_from_slice(&encrypt_priv_key.0)?,
+                )?
+                .into();
+                let aug_factor = AugmentationFactor::generate_new(recrypt);
+                // TODO reviewers: should retry augfactor internally?
```
BobWall23: I think so

src/lib.rs
```diff
@@ -90,41 +96,107 @@ pub struct IronOxide {
...
+/// for private key rotation.
+pub fn initialize_check_rotation(device_context: &DeviceContext) -> Result<InitAndRotationCheck> {
+    Runtime::new().unwrap().block_on(
+        internal::user_api::user_get_current(device_context.auth()).and_then(|curr_user| {
+            let ironoxide = IronOxide::create_with_side_effects(&curr_user, &device_context);
+
+            if curr_user.needs_rotation() {
+                Ok(InitAndRotationCheck::RotationNeeded(
+                    ironoxide,
+                    PrivateKeyRotationCheckResult {
```
BobWall23: What if any groups need rotation?

src/internal/mod.rs
clintfred: I agree. See https://github.com/IronCoreLabs/recrypt-rs/issues/91

src/lib.rs
clintfred: That capability will not exist until https://github.com/IronCoreLabs/ironoxide/issues/45 is complete.

src/internal/user_api/mod.rs
clintfred: I did one internal retry.

src/internal/user_api/requests.rs
```diff
@@ -121,6 +131,88 @@ pub mod user_verify {
...
+        current_key_id: u64,
+        user_private_key: EncryptedPrivateKey,
+        needs_rotation: bool,
+    }
+
+    impl From<UserUpdatePrivateKeyResponse> for UserUpdatePrivateKeyResult {
+        fn from(resp: UserUpdatePrivateKeyResponse) -> Self {
+            UserUpdatePrivateKeyResult {
+                user_key_id: resp.current_key_id,
+                user_master_private_key: internal::user_api::EncryptedPrivateKey(
```
coltfred: A from instance would make this more concise.

src/lib.rs
```diff
@@ -90,41 +100,107 @@ pub struct IronOxide {
...
+    /// Initialization succeeded, and no requests for private key rotations were present
+    NoRotationNeeded(IronOxide),
+    /// Initialization succeeded, but some keys should be rotated
+    RotationNeeded(IronOxide, PrivateKeyRotationCheckResult),
+}
+
+impl InitAndRotationCheck {
+    /// Caller asked to check rotation on initialize, but doesn't want to handle the result.
+    /// Consider using [initialize](fn.initialize.html) instead.
+    pub fn unwrap(self) -> IronOxide {
```
coltfred: I thought we'd decided on a different name for this, but can't think of it right now. `unwrap` in other apis will throw if it's not the "good" value so it seems like a bad name for this context.

src/lib.rs
```diff
@@ -90,41 +100,107 @@ pub struct IronOxide {
...
+
+impl PrivateKeyRotationCheckResult {
+    pub fn user_rotation_needed(&self) -> Option<UserId> {
+        match &self.rotations_needed {
+            EitherOrBoth::Left(u) | EitherOrBoth::Both(u, _) => Some(u.clone()),
+            _ => None,
+        }
+    }
+
+    //    pub fn group_rotation_needed() -> Option<Vec<GroupId>> {
```
coltfred: I think we should remove these if you're not planning to deal with them now.

src/lib.rs
```diff
@@ -90,41 +100,107 @@ pub struct IronOxide {
...
-                user_master_pub_key: current_user_public_key,
-                rng: Mutex::new(ReseedingRng::new(
-                    rand_chacha::ChaChaCore::from_entropy(),
-                    BYTES_BEFORE_RESEEDING,
-                    EntropyRng::new(),
-                )),
-            })
-            .ok_or(IronOxideErr::InitializeError)
+    .and_then(|current_user| {
+        Ok(IronOxide::create_with_side_effects(
```
coltfred: Doesn't seem like you should need the Ok here. and_then just needs something that can go to future and I think a bare value implements that.

src/lib.rs
```diff
@@ -90,41 +100,107 @@ pub struct IronOxide {
...
+    )
 }
 
 impl IronOxide {
     /// Get the `DeviceContext` instance that was used to create this SDK instance
     pub fn device(&self) -> &DeviceContext {
         &self.device
     }
+
+    fn create_with_side_effects(
```
coltfred: I don't understand this name.

tests/user_ops.rs
```diff
@@ -83,12 +84,79 @@ fn user_create_good_with_devices() {
...
+
+    let encrypt_result = sdk.document_encrypt(
+        &bytes,
+        &DocumentEncryptOpts::with_explicit_grants(None, None, true, vec![]),
+    )?;
+    let encrypted_data = encrypt_result.encrypted_data();
+
+    //rotate the private key
+    let rotation_result = sdk.user_rotate_private_key(common::USER_PASSWORD)?;
+    dbg!(&rotation_result);
```
coltfred: Remove?

src/internal/user_api/mod.rs
```diff
@@ -236,6 +237,115 @@ pub fn user_create<CR: rand::CryptoRng + rand::RngCore>(
...
+                )
+                .map(|resp| resp.into())
+            },
+        )
+}
+
+/// Generate a new private key and augmentation factor from a given PrivateKey.
+/// There is a very small chance that an augmentation factor could not be compatible with
+/// the given PrivateKey, so we retry once internally before giving the caller an error.
+fn generate_private_key_and_augmentation_factor<R: KeyGenOps>(
```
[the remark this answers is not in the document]
clintfred: I agree that this name isn't very good. I've moved this code to internal/mod.rs as it will be needed for group private key rotation (I think) and renamed it as `augment_private_key_with_retry`

src/internal/user_api/mod.rs
```diff
@@ -498,4 +616,75 @@ mod test {
...
+            _to_public_key: &RePublicKey,
+            _signing_keypair: &SigningKeypair,
+        ) -> Result<recrypt::api::TransformKey, RecryptErr> {
+            unimplemented!()
+        }
+    }
+    #[test]
+    fn gen_private_key_and_augmentation_factor_retries_once() {
+        let recrypt_mock = MockKeyGenOps::default();
+        let good_re_private_key = RePrivateKey::new([100u8; 32]); // good private key
```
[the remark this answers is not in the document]
clintfred: great suggestion. Done.

src/internal/user_api/requests.rs
```diff
@@ -121,6 +131,88 @@ pub mod user_verify {
...
+        pub(in crate::internal) needs_rotation: bool,
+        pub(in crate::internal) groups_needing_rotation: Vec<String>,
+    }
+
+    pub fn get_curr_user(
+        auth: &RequestAuth,
+    ) -> impl Future<Item = CurrentUserResponse, Error = IronOxideErr> + '_ {
+        auth.request.get(
+            "users/current",
+            RequestErrorCode::EdekTransform,
```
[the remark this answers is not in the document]
clintfred: :eyes: totally wrong; thanks

src/lib.rs
clintfred: I _think_ I remember that conversation...
Potential names might be
`discard_result()`
`skip_check()`

src/lib.rs
clintfred: Hrm. A couple of things here. I don't think bare values implement `IntoFuture`, and this isn't actually a `Future` at all here, it's a `Result`. So `map` can be used to get rid of the `Ok`. I did that.

src/lib.rs
clintfred: Meant to indicate that we are depending on RNG seeding to create this IronOxide. Could potentially be shortened to `create` -- `new` seems to have the connotation (to me) that the construction is pure.

src/lib.rs
clintfred: I'm inclined to leave them here as we will be tackling this right away and otherwise we'd need to move the ideas here to tickets or elsewhere.

src/internal/mod.rs
```diff
@@ -477,6 +485,51 @@ impl<'de> Deserialize<'de> for PrivateKey {
...
+            } else {
+                Ok(augmented_key.into())
+            }
+        }
+    }
+}
+
+/// Private key used to augment another PrivateKey
+#[derive(Clone, Debug)]
+pub(crate) struct AugmentationFactor(PrivateKey);
```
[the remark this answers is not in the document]
clintfred: Perhaps I'm missing the point, but it seems like for us to realize any benefit from implementing `Zero` and `PartialEq` for `AugmentationFactor` we would also need to implement those on `internal::PrivateKey`, which we have avoided so far.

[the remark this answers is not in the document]
coltfred: Ok, I'm not sure we should actually care about this in general, but you're right that it's the way we've been doing it.

src/lib.rs
coltfred: How about `discard_check`? (I'm not trying to troll)

src/lib.rs
coltfred: I hate leaving things laying around the codebase (especially when we could shift any day). The code here doesn't feel like it really provides _that_ much value....

src/lib.rs
coltfred: create with a comment seems better to me.

src/lib.rs
clintfred: or how about `skip_result`!? `discard_check` actually seems pretty good.

src/internal/mod.rs
```diff
@@ -477,6 +486,51 @@ impl<'de> Deserialize<'de> for PrivateKey {
...
+    /// Augment this private key with another, producing a new PrivateKey
+    fn augment(&self, augmenting_key: &AugmentationFactor) -> Result<PrivateKey, IronOxideErr> {
+        use recrypt::Revealed;
+        let zero: RecryptPrivateKey = RecryptPrivateKey::new([0u8; 32]);
+        if Revealed((augmenting_key.clone().0).0) == Revealed(zero.clone()) {
+            Err(IronOxideErr::UserPrivateKeyRotationError(
+                "Augmenting key cannot be zero".into(),
+            ))
+        } else {
+            // this subtraction needs to be the additive inverse of what the service is doing
```
BobWall23: I thought I had mentioned this, but maybe that was somewhere else. Instead of doing the subtraction then comparing the result to zero, you could compare the augmenting key with self.0. If they are equal, it will end up failing with a zero value.

src/internal/user_api/requests.rs
```diff
@@ -121,6 +137,86 @@ pub mod user_verify {
...
+/// PUT /users/{userId}/keys/{userKeyId}
+pub mod user_update_private_key {
+    use super::*;
+    use crate::internal::user_api::UserUpdatePrivateKeyResult;
+
+    #[derive(Serialize, Debug)]
+    #[serde(rename_all = "camelCase")]
+    pub struct UserUpdatePrivateKey {
+        user_private_key: EncryptedPrivateKey,
+        augmentation_factor: AugmentationFactor,
```
BobWall23: I'm not seeing the Schnorr signature - is it wrapped up in the `EncryptedPrivateKey`?

clintfred: Should user private key rotation use a Schnorr signature? As of now we don't. Since this call is made with a device it doesn't seem necessary to me at first glance.

src/internal/mod.rs
clintfred: This is somewhat worse now as it it will require another clone of the `self` value, but it allows eliminating a level of nesting which is nice. The performance impact of the clone will go away when we fix Recrypt's `Revealed` type.

BobWall23: I think it's good to eliminate the subtraction if it is easy to do.

src/internal/user_api/requests.rs
BobWall23: My mistake - the thing we sign with Schnorr is in the device transform key, to make sure that was generated with the updated private key.

src/internal/mod.rs
clintfred: Maybe I'm still not understanding what you are proposing. The subtraction isn't eliminated it's simply delayed. If the augmenting value were the same as self, it would be eliminated, but that is expected to be incredibly rare.

See https://github.com/IronCoreLabs/ironoxide/pull/59/files#diff-b17bedabf75c7285b05afbf395928056R491 for my implementation of delaying the subtraction.

BobWall23: You're right. Working to eliminate something that will probably never happen was not worth thinking about.

PR #67, opened 2019-11-05
Add signature to the add admin request

coltfred:
see #48 

TODO
- [x] Ensure we don't have more tests we should/could do.

### DIFF
--- a/src/internal/group_api/mod.rs
+++ b/src/internal/group_api/mod.rs
@@ -4,8 +4,8 @@ use crate::{
         group_api::requests::{group_list::GroupListResponse, GroupUserEditResponse},
         rest::json::TransformedEncryptedValue,
         user_api::{self, UserId},
-        validate_id, validate_name, IronOxideErr, PrivateKey, PublicKey, RequestAuth, TransformKey,
-        WithKey,
+        validate_id, validate_name, IronOxideErr, PrivateKey, PublicKey, RequestAuth,
+        SchnorrSignature, TransformKey, WithKey,
     },
 };
 use chrono::{DateTime, Utc};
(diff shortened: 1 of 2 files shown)


src/internal/group_api/mod.rs
```diff
@@ -445,6 +450,12 @@ pub fn group_add_admins<'a, CR: rand::CryptoRng + rand::RngCore>(
                 encrypted_group_key.try_into()?,
                 &device_private_key.recrypt_key(),
             )?;
+            let private_group_key = recrypt.derive_private_key(&plaintext);
+            let recrypt_schnorr_sig = recrypt.schnorr_sign(
+                &private_group_key,
+                &group_get.group_master_public_key.into(),
+                group_id,
+            );
```
clintfred: Maybe you feels that it's too simple to test, but you could pull this out into a function and we could have a test that shows that this signature is over (exactly) group id...

coltfred: I feel like because of our nice types that is very obvious here and isn't made more obvious by refactoring. I did try and do the refactoring, I just don't think it makes things more clear.

clintfred: Mostly playing devil's advocate here: I'm not saying it isn't obvious or correct. I'm saying you can't write a unit test for the signature with it buried in this larger function. If someone changed the signature to be over `(group_id, group_id)` there wouldn't be a unit test to catch that.

This case is very simple, but in general it's nice to have tests on signatures.

Still approved, btw. :)

coltfred: Right. If that happened, the server would reject it (integration tests would fail). I did add a test to ensure the sanity of the hashable impl on GroupId, though.

Issue #64, opened 2019-11-05
Create Group with needs_rotation


clintfred:
As an application developer
I want to be able to create a cryptographic group before the admins and members have taken ownership of their keys
so groups can be created independently of the admins and members logging in and initializing the SDK

AC:
* [ ] `GroupCreateOpts` should contain a `needs_rotation` option
* [ ] Result coming back from group create should contain the `needs_rotation` flag

PR #68, opened 2019-11-05
[#64] Adds needs_rotation to group creation

clintfred:
see #64 

Adds need_rotation to `GroupCreateOpts`, allowing a group to be created with its private key marked for rotation


clintfred:
Did create
* https://github.com/IronCoreLabs/ironoxide-scala/issues/33
* https://github.com/IronCoreLabs/ironoxide-java/issues/60

PR #69, opened 2019-11-05
Make test environment configurable

giarc3:
TODO:
- [x] Update the CHANGELOG.md
- [x] Update the documentation for the latest design changes
- [x] Verify that Travis is decrypting successfully


tests/INTEGRATION-TESTS.md
```diff
@@ -1,24 +1,50 @@
...
+## Setting Up Integration Tests
 
-Integration tests are run as part of a PR build on Travis. These keys are stored as a Travis secret.
+In order to run the integration tests, you must provide IronOxide with an Identity Assertion Key file, an IronCore Config file, and the URL you would like to test against. This will require you to create a project, segment, and Identity Assertion Key using the admin console interface.
 
-The integration test run against IronCore's staging environment and require some tests keys. These can be found in `tests/testkeys/rsa_private.pem.iron`. _Currently only IronCore devs have access to these keys._ The following ironhide command will decrypt the developer test keys. 
+### Identity Assertion Key File
 
-`$ ironhide file:decrypt rsa_private.pem.iron`
+This file must be downloaded from the admin console interface immediately after creating a new Identity Assertion Key. It should be named `rsa_private.pem` and placed in `./tests/testkeys`.
```
coltfred: This name is no longer correct.

tests/INTEGRATION-TESTS.md
```diff
@@ -1,24 +1,50 @@
...
+
+### IronCore Config File
+
+This file can be downloaded from the admin console on creation of the very first project. For subsequent projects, it will need to be created manually. The file is of the form:
+
+```json
+{
+  "projectId": { YOUR_PROJECT_ID },
```
coltfred: The spacing here seems weird. If I didn't know better I'd think you wanted an object as the value for `projectId`. I think removing the `{` and just having the capital snake case place holder is easier to read.

tests/INTEGRATION-TESTS.md
```diff
@@ -1,24 +1,50 @@
...
+}
+```
+
+Note that case is significant for the key names.
+
+This file must be named `ironcore-config.json` and placed in `./tests/testkeys`.
+
+### Environment URL
+
+By default, IronOxide will test against the staging build, but it can also test against dev, prod, or any other environment. This is specified with the environment variable `IRONCORE_ENV`, which can be set before running `cargo t`. There are several built-in environment URLs, or one can be specified. To do this, run the following:
```
coltfred: "staging build" would be better said as "staging environment". Also, this currently isn't true as it'd run against production. We'll discuss the ramifications of that as a team tomorrow morning.

tests/INTEGRATION-TESTS.md
giarc3: Fixed

tests/INTEGRATION-TESTS.md
giarc3: Changed. Please double-check that you think it's clear now.

tests/INTEGRATION-TESTS.md
giarc3: The document has been re-worked. Let me know if you have more suggestions.

PR #72, opened 2019-11-08
Add members on group create

giarc3:
- Made `GroupCreateOpts` contain a list of `UserIds` that should be added as members to the group. `group_create()` will fail if one of these `UserIds` cannot be found.
- Made `group_create()` return a `GroupCreateApiResponse` with a `try_into()` for a new `GroupCreateResult`. This result is similar to `GroupMetaResult` but includes a list of group admins and members.
- Added a test for adding members on create with the `members` list.
- Changed some tests to return Results and use `?` instead of `unwrap()` to get better error messages.

TODO:
- [x] Changelog
- [x] Discuss TODO questions aimed at reviewers

### DIFF
--- a/src/group.rs
+++ b/src/group.rs
@@ -55,6 +55,20 @@ impl GroupCreateOpts {
             needs_rotation,
         }
     }
+
+    fn standardize(self, calling_id: &UserId) -> GroupCreateOpts {
+        let mut new_members = self.members.clone();
+        if self.add_as_member && !self.members.contains(calling_id) {
+            new_members.push(calling_id.clone());
+        }
+        GroupCreateOpts::new(
+            self.id,
+            self.name,
+            self.add_as_member,
+            new_members,
+            self.needs_rotation,
+        )
+    }
 }
 
 impl Default for GroupCreateOpts {
@@ -166,17 +180,14 @@ impl GroupOps for crate::IronOxide {
 
     fn group_create(&self, opts: &GroupCreateOpts) -> Result<GroupCreateResult> {
         let mut rt = Runtime::new().unwrap();
+
         let GroupCreateOpts {
             id: maybe_id,
             name: maybe_name,
-            add_as_member,
-            mut members,
+            add_as_member: _,
+            members,
             needs_rotation,
-        } = opts.clone();
-
-        if add_as_member {
-            members.push(self.device.auth().account_id().clone());
-        };
+        } = opts.clone().standardize(self.device.auth().account_id());
 
         rt.block_on(group_api::group_create(
             &self.recrypt,
@@ -184,7 +195,6 @@ impl GroupOps for crate::IronOxide {
             &self.user_master_pub_key,
             maybe_id,
             maybe_name,
-            add_as_member,
             &members,
             needs_rotation,
         ))
--- a/src/group.rs
+++ b/src/group.rs
@@ -57,15 +57,18 @@ impl GroupCreateOpts {
     }
 
     fn standardize(self, calling_id: &UserId) -> GroupCreateOpts {
-        let mut new_members = self.members.clone();
-        if self.add_as_member && !self.members.contains(calling_id) {
-            new_members.push(calling_id.clone());
-        }
+        let standardized_members = if self.add_as_member && !self.members.contains(calling_id) {
+            let mut members = self.members.clone();
+            members.push(calling_id.clone());
+            members
+        } else {
+            self.members
+        };
         GroupCreateOpts::new(
             self.id,
             self.name,
             self.add_as_member,
-            new_members,
+            standardized_members,
             self.needs_rotation,
         )
     }
--- a/src/group.rs
+++ b/src/group.rs
@@ -1,6 +1,6 @@
 pub use crate::internal::group_api::{
-    GroupAccessEditErr, GroupAccessEditResult, GroupGetResult, GroupId, GroupListResult,
-    GroupMetaResult, GroupName,
+    GroupAccessEditErr, GroupAccessEditResult, GroupCreateResult, GroupGetResult, GroupId,
+    GroupListResult, GroupMetaResult, GroupName,
 };
 use crate::{
     internal::{group_api, user_api::UserId},
@@ -18,6 +18,9 @@ pub struct GroupCreateOpts {
     // true (default) - creating user will be added to the group's membership (in addition to being the group's admin);
     // false - creating user will not be added to the group's membership
     add_as_member: bool,
+    // list of users to add as members to the group.
+    // note: even if `add_as_member` is false, the calling user will be added as a member if they are in this list.
+    members: Vec<UserId>,
     // true - group's private key will be marked for rotation
     // false (default) - group's private key will not be marked for rotation
     needs_rotation: bool,
@@ -39,12 +42,14 @@ impl GroupCreateOpts {
         id: Option<GroupId>,
         name: Option<GroupName>,
         add_as_member: bool,
+        members: Vec<UserId>,
         needs_rotation: bool,
     ) -> GroupCreateOpts {
         GroupCreateOpts {
             id,
             name,
             add_as_member,
+            members,
             needs_rotation,
         }
     }
@@ -53,7 +58,7 @@ impl GroupCreateOpts {
 impl Default for GroupCreateOpts {
     fn default() -> Self {
         // membership is the default!
-        GroupCreateOpts::new(None, None, true, false)
+        GroupCreateOpts::new(None, None, true, Vec::new(), false)
     }
 }
 
@@ -68,7 +73,7 @@ pub trait GroupOps {
     ///
     /// # Arguments
     /// `group_create_opts` - See `GroupCreateOpts`. Use the `Default` implementation for defaults.
-    fn group_create(&self, group_create_opts: &GroupCreateOpts) -> Result<GroupMetaResult>;
+    fn group_create(&self, group_create_opts: &GroupCreateOpts) -> Result<GroupCreateResult>;
 
     /// Get the full metadata for a specific group given its ID.
     ///
@@ -157,14 +162,26 @@ impl GroupOps for crate::IronOxide {
         rt.block_on(group_api::list(self.device.auth(), None))
     }
 
-    fn group_create(&self, opts: &GroupCreateOpts) -> Result<GroupMetaResult> {
+    fn group_create(&self, opts: &GroupCreateOpts) -> Result<GroupCreateResult> {
         let mut rt = Runtime::new().unwrap();
         let GroupCreateOpts {
             id: maybe_id,
             name: maybe_name,
             add_as_member,
+            members,
             needs_rotation,
         } = opts.clone();
+
+        let modified_add_as_member;
+        if members.contains(self.device.auth().account_id()) {
+            modified_add_as_member = true;
+        } else {
+            modified_add_as_member = add_as_member;
+        }
+        let modified_members: Vec<UserId> = members
+            .into_iter()
+            .filter(|id| id.id() != self.device.auth().account_id().id())
+            .collect();
 
         rt.block_on(group_api::group_create(
             &self.recrypt,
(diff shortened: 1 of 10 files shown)


src/internal/group_api/mod.rs
```diff
@@ -307,37 +361,79 @@ pub fn group_create<'a, CR: rand::CryptoRng + rand::RngCore>(
...
+                                })
+                                .collect();
+
+                        let mut member_info = maybe_member_info?;
+
+                        if add_as_member {
+                            // TODO (question for reviewers): is it better to have this code duplication here,
+                            // or to add the calling UserId to members list earlier and just have the public key
+                            // looked up with all the others (even though it was already passed in) to make
+                            // it cleaner?
```
ToniaSchlick: I think the way you have it is better then doing an unnecessary look-up.

src/group.rs
```diff
@@ -157,22 +162,35 @@ impl GroupOps for crate::IronOxide {
...
         let mut rt = Runtime::new().unwrap();
         let GroupCreateOpts {
             id: maybe_id,
             name: maybe_name,
             add_as_member,
+            members,
             needs_rotation,
         } = opts.clone();
 
+        let modified_add_as_member;
```
coltfred: You should be able to assign the if/else to this let.

src/group.rs
```diff
@@ -157,22 +162,35 @@ impl GroupOps for crate::IronOxide {
         rt.block_on(group_api::list(self.device.auth(), None))
     }
 
-    fn group_create(&self, opts: &GroupCreateOpts) -> Result<GroupMetaResult> {
+    fn group_create(&self, opts: &GroupCreateOpts) -> Result<GroupCreateResult> {
```
coltfred: We don't usually do any business logic up in this layer of the api so this logic should go down into group_api::group_create instead of being done up here.

src/group.rs
```diff
@@ -157,22 +162,35 @@ impl GroupOps for crate::IronOxide {
...
         let GroupCreateOpts {
             id: maybe_id,
             name: maybe_name,
             add_as_member,
+            members,
             needs_rotation,
         } = opts.clone();
 
+        let modified_add_as_member;
+        if members.contains(self.device.auth().account_id()) {
```
coltfred: `let modified_add_as_member = add_as_member || members.contains(self.device.auth().account_id())` is an easier way to write this.

src/group.rs
```diff
@@ -157,22 +162,35 @@ impl GroupOps for crate::IronOxide {
...
+            modified_add_as_member = true;
+        } else {
+            modified_add_as_member = add_as_member;
+        }
+        let modified_members: Vec<UserId> = members
+            .into_iter()
+            .filter(|id| id.id() != self.device.auth().account_id().id())
```
coltfred: `.filter(|id| id != self.device.auth().account_id())` -- You can compare using the id like this instead.

src/internal/group_api/mod.rs
```diff
@@ -136,6 +136,57 @@ impl GroupMetaResult {
...
+    group_master_public_key: PublicKey,
+    is_admin: bool,
+    is_member: bool,
+    admins: Vec<UserId>,
+    members: Vec<UserId>,
+    created: DateTime<Utc>,
+    updated: DateTime<Utc>,
+    needs_rotation: Option<bool>,
+}
+impl GroupCreateResult {
```
coltfred: I assume this is mostly copy, but we should comment all of these instead of just some of them.

src/group.rs
```diff
@@ -157,22 +162,35 @@ impl GroupOps for crate::IronOxide {
...
             needs_rotation,
         } = opts.clone();
 
+        let modified_add_as_member;
+        if members.contains(self.device.auth().account_id()) {
+            modified_add_as_member = true;
+        } else {
+            modified_add_as_member = add_as_member;
+        }
+        let modified_members: Vec<UserId> = members
```
coltfred: You already know if the member is in the list (because of the contains check). It seems like we could avoid this filter in that case. I'm happy to help you write that code so we can make things more clear if you want.

src/internal/group_api/mod.rs
```diff
@@ -307,37 +360,78 @@ pub fn group_create<'a, CR: rand::CryptoRng + rand::RngCore>(
...
+        .and_then(move |member_ids_and_keys| {
+            // this will occur when one of the UserIds in the members list cannot be found
+            if member_ids_and_keys.len() != members.len() {
+                // figure out which user ids could not be found in the database and include them in the error message.
+                use std::{collections::HashSet, iter::FromIterator};
+                let desired_members_set: HashSet<&UserId> = HashSet::from_iter(members);
+                let found_members: Vec<UserId> =
+                    member_ids_and_keys.into_iter().map(|(x, _)| x).collect();
+                let found_members_set: HashSet<&UserId> = HashSet::from_iter(&found_members);
+                // TODO (reviewers): this double reference scares me, but seems to work? Is there a better way to do this?
```
coltfred: I don't have a better way to do it.

src/internal/group_api/mod.rs
```diff
@@ -307,37 +360,78 @@ pub fn group_create<'a, CR: rand::CryptoRng + rand::RngCore>(
...
+                                    );
+                                    maybe_transform_key.map(|transform_key| {
+                                        (id, member_pub_key, transform_key.into())
+                                    })
+                                })
+                                .collect();
+
+                        let mut member_info = maybe_member_info?;
+
+                        if add_as_member {
```
coltfred: It seems like you could just add the id and keys to the members list to avoid this duplication.

src/group.rs
giarc3: As we talked about, I'll need some help getting the logic into the other layer

src/internal/group_api/mod.rs
giarc3: Added comments to all

src/group.rs
giarc3: I've changed this section, but it's good to know that I can compare the UserIds directly

src/internal/group_api/mod.rs
giarc3: Great, then I'll try to be less scared. :ghost:

src/internal/group_api/mod.rs
giarc3: I'm having issues modifying references to vectors in this function without the compiler eventually yelling at me, so this was my way to avoid looking up the public key with the others since we already know it. Hopefully we'll get the compiler sorted out.

src/group.rs
```diff
@@ -32,28 +35,46 @@ impl GroupCreateOpts {
...
             id,
             name,
             add_as_member,
+            members,
             needs_rotation,
         }
     }
+
+    fn standardize(self, calling_id: &UserId) -> GroupCreateOpts {
+        let mut new_members = self.members.clone();
```
coltfred: You might be able to do this clone inside the  if to make it not need to happen in the case where we don't have to add the calling users id.

src/internal/group_api/mod.rs
```diff
@@ -297,47 +359,81 @@ pub(crate) fn get_group_keys<'a>(
...
+                                    })
+                                } else {
+                                    Ok((id, (user_pub_key, None)))
+                                }
+                            })
+                            .collect();
+
+                        let user_info = maybe_user_info?;
+
+                        let maybe_user_info = if user_info.len() != 0 {
```
coltfred: Reusing this variable name seems confusing to me. Maybe name the first one `error_or_user_info` or something.

src/internal/group_api/requests.rs
```diff
@@ -194,31 +229,36 @@ pub mod group_create {
...
         EncryptedOnceValue::try_from(re_encrypted_once_value)
             .into_future()
             .and_then(move |enc_msg| {
+                // this currently only holds members, will need to expand for admins later
+                let req_members = user_info.map(|member| {
+                    member
+                        .into_iter()
+                        .map(|(mem_id, (pub_key, trans_key))| GroupMember {
+                            user_id: mem_id,
+                            transform_key: trans_key.unwrap().into(), // we can unwrap because we know all members had it calculated
```
coltfred: I would rather match on the `Some(trans_key)` and leave the `None` case empty (it'll be used in the next PR).

src/internal/group_api/mod.rs
giarc3: I agreed and just changed that

src/group.rs
giarc3: I made this a bit uglier but hopefully also better

src/internal/group_api/requests.rs
giarc3: changed to map over the option then flatten out the Nones

src/group.rs
```diff
@@ -32,28 +35,49 @@ impl GroupCreateOpts {
...
+            self.needs_rotation,
+        )
+    }
 }
 
 impl Default for GroupCreateOpts {
     fn default() -> Self {
         // membership is the default!
-        GroupCreateOpts::new(None, None, true, false)
+        GroupCreateOpts::new(None, None, true, Vec::new(), false)
```
clintfred: I would suggest creating a new vec with `vec![]`
Applicable in some other places in this PR as well...

src/internal/group_api/mod.rs
```diff
@@ -297,47 +359,75 @@ pub(crate) fn get_group_keys<'a>(
...
+                        // TransformKey is optional because we need it for members, but won't need it
+                        // when we expand this to include admins
+                        let member_info_result: Result<
+                            HashMap<UserId, (PublicKey, Option<TransformKey>)>,
+                            _,
+                        > = member_ids_and_keys
+                            .into_iter()
+                            .map(|(id, user_pub_key)| {
+                                let maybe_transform_key = recrypt.generate_transform_key(
+                                    &group_priv_key.clone().into(),
```
clintfred: is this clone needed? I don't see this being used below...

giarc3: Removing it gives
```cannot move out of `group_priv_key`, a captured variable in an `FnMut` closure```
```move occurs because `group_priv_key` has type `recrypt::api::PrivateKey`, which does not implement the `Copy` trait rustc(E0507)```

File: src/internal/group_api/mod.rs [hunk 15]
```diff
@@ -297,47 +359,75 @@ pub(crate) fn get_group_keys<'a>(
 /// `user_master_pub_key` - public key of the user creating this group.
 /// `group_id` - unique id for the group within the segment.
 /// `name` - name for the group. Does not need to be unique.
-/// `add_as_member` - if true the user represented by the current DeviceContext will also be added to the group's membership.
-///     If false, the user will not be an member (but will still be an admin)
+/// `members` - list of user ids to add as members of the group. This list takes priority over `add_as_member`,
```
clintfred: I don't understand the reference to `add_as_member` in this comment as it is not part of this function's signature.

src/internal/group_api/mod.rs
```diff
@@ -297,47 +359,75 @@ pub(crate) fn get_group_keys<'a>(
...
-                } else {
-                    None
-                };
-                (group_pub_key, encrypted_group_key, transform_key)
-            })
+) -> impl Future<Item = GroupCreateResult, Error = IronOxideErr> + 'a {
+    user_api::user_key_list(auth, members)
+        .and_then(move |member_ids_and_keys| {
+            // this will occur when one of the UserIds in the members list cannot be found
+            if member_ids_and_keys.len() != members.len() {
```
clintfred: Maybe this branch should be its own function? Would this be reusable when adding admins? It certainly would be more testable on its own, IMO.

src/internal/group_api/requests.rs
```diff
@@ -94,6 +96,39 @@ impl TryFrom<GroupGetApiResponse> for GroupGetResult {
     }
 }
 
+#[derive(Deserialize, Debug, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct GroupCreateApiResponse {
+    pub(crate) id: GroupId,
```
clintfred: I have been making these response objects `pub(in crate:internal)` recently as we should never need these outside of the `internal` portion of the crate.

Issue #58, opened 2019-10-22
All public types should provide Clone, Debug, Hash, and PartialEq


clintfred:
There is inconsistency among the public Result types as to if they provide `Debug`, `Clone`, and `PartialEq`. Unless there's a good reason, all of these types should provide instances these traits.

clintfred:
Perhaps we should also provide Serialize and Deserialize? I think the only reason to not do that is that our public API would depend on serde, which may be undesirable in some "minimal deps" usecases.

Issue #18, opened 2019-07-14
Bad error in the case of policy that evaluates to a group that doesn't exist


coltfred:
There are other sets of steps which produce this same result, but this is the simplest IMO.

1. Initialize IronOxide.
2. Call document_encrypt with `DocumentEncryptOpts` which only has `ExplicitGrants{grant_to_author:false, grants: [UserOrGroup::User("does_not_exist")}`

Result: The function will error with `grants' failed validation with the error 'Access must be granted to document DocumentId("XXXXXXXXXXXXXXXX") by explicit grant or via a policy'` 

Expected result: The function should error telling you which people it tried to share with and why that didn't work. It's right to error all the way out, but since I did send in a valid ExplicitGrant I shouldn't get this error. As the caller I'm confused because I did send an ExplicitGrant, it's just a grant that didn't result in valid users.

I think we should add an error that catches the case and tells them what users or groups we tried to share with and why it couldn't be successful.

clintfred:
As of #27 the code to resolve public keys from grants/policies was consolidated in `resolve_keys_for_grants`. This would be a perfect place to use a non-empty list like [vec1](https://github.com/dac-gmbh/vec1)

`resolve_keys_for_grants` should with return a `vec1` or error, preventing the confusing error situation described above.

PR #76, opened 2019-11-20
Add admins/owner on group create

giarc3:
- Added `add_as_admin`, `admins`, and `owner` to `GroupCreateOpts`.
- Made `GroupCreateOpts::standardize()` return a new `GroupCreateOptsStd` that is destructured by `group_create`.
- Added additional validation with combinations of `owner` and `admins`.
- Added `owner` to group response structs.
- Added tests for adding admins/owner on group create.
- Restructured lots of tests to return Result and use `?` because better error messages were needed during testing.

TODO:
- [x] Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 0.14.0 (Unreleased)
 
+- [[#72](https://github.com/IronCoreLabs/ironoxide/pull/72)]
+  - Allows adding members at group creation time
 - [[#69](https://github.com/IronCoreLabs/ironoxide/pull/69)]
   - Allows changing of IronCore environment at runtime.
 - [[#64](https://github.com/IronCoreLabs/ironoxide/pull/64)]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,11 @@
 
 ## 0.14.0 (Unreleased)
 
+- [[#76](https://github.com/IronCoreLabs/ironoxide/pull/76)]
+  - Allows adding admins at group creation time.
+  - Allow specifying an owner at group creation time.
 - [[#72](https://github.com/IronCoreLabs/ironoxide/pull/72)]
-  - Allows adding members at group creation time
+  - Allows adding members at group creation time.
 - [[#69](https://github.com/IronCoreLabs/ironoxide/pull/69)]
   - Allows changing of IronCore environment at runtime.
 - [[#64](https://github.com/IronCoreLabs/ironoxide/pull/64)]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - [[#76](https://github.com/IronCoreLabs/ironoxide/pull/76)]
   - Allows adding admins at group creation time.
-  - Allow specifying an owner at group creation time.
+  - Allows specifying an owner at group creation time.
 - [[#72](https://github.com/IronCoreLabs/ironoxide/pull/72)]
   - Allows adding members at group creation time.
 - [[#69](https://github.com/IronCoreLabs/ironoxide/pull/69)]
--- a/src/group.rs
+++ b/src/group.rs
@@ -6,24 +6,40 @@ use crate::{
     internal::{group_api, user_api::UserId},
     Result,
 };
+use std::{collections::HashSet, iter::FromIterator};
 use tokio::runtime::current_thread::Runtime;
 
+struct GroupCreateOptsStd {
+    id: Option<GroupId>,
+    name: Option<GroupName>,
+    owner: Option<UserId>,
+    admins: Vec<UserId>,
+    members: Vec<UserId>,
+    users: Vec<UserId>,
+    needs_rotation: bool,
+}
+
 #[derive(Clone)]
-// TODO: docs
 /// Options for group creation.
 pub struct GroupCreateOpts {
     // unique id of a group within a segment. If none, the server will assign an id.
     id: Option<GroupId>,
     // human readable name of the group. Does not need to be unique.
     name: Option<GroupName>,
-
+    // true (default) - creating user will be added as an admin of the group.
+    // false - creating user will not be added as an admin of the group.
+    add_as_admin: bool,
     // true (default) - creating user will be added to the group's membership (in addition to being the group's admin);
     // false - creating user will not be added to the group's membership
-    add_as_admin: bool,
     add_as_member: bool,
+    // None (default) - The creating user will be the owner of the group.
+    // Some(UserId) - The provided user will be the owner of the group.
+    // Note that the owner must be included in the `admins` list.
     owner: Option<UserId>,
+    // list of users to add as admins of the group
+    // note: even if `add_as_admin` is false, the calling user will be added as an admin if they are in this list.
     admins: Vec<UserId>,
-    // list of users to add as members to the group.
+    // list of users to add as members of the group.
     // note: even if `add_as_member` is false, the calling user will be added as a member if they are in this list.
     members: Vec<UserId>,
     // true - group's private key will be marked for rotation
@@ -38,16 +54,18 @@ impl GroupCreateOpts {
     /// - `id` - Unique id of a group within a segment. If none, the server will assign an id.
     /// - `name` - Human readable name of the group. Does not need to be unique. Will **not** be encrypted.
     /// - `add_as_admin`
-    ///     - `true` - The creating user will be added as an admin of the group.
-    ///     - `false` - The creating user will not be an admin of the group.
+    ///     - true (default) - The creating user will be added as an admin of the group.
+    ///     - false - The creating user will not be an admin of the group.
     /// - `add_as_member`
-    ///     - `true` - The creating user will be added as a member of the group.
-    ///     - `false` - The creating user will not be a member of the group.
-    /// - `owner`
-    ///     - `Some(UserId)` - The provided user will be the owner of the group.
-    ///     - `None` - The creating user will be the owner of the group.
-    /// - `admins` - List of users to be added as admins of the group.
-    /// - `members` - List of users to be added as members of the group.
+    ///     - true (default) - The creating user will be added as a member of the group.
+    ///     - false - The creating user will not be a member of the group.
+    /// - `owner` - Note that the owner must be included in the `admins` list.
+    ///     - None (default) - The creating user will be the owner of the group.
+    ///     - Some(UserId) - The provided user will be the owner of the group.
+    /// - `admins` - List of users to be added as admins of the group. This list takes priority over `add_as_admin`,
+    ///             so the calling user will be added as a member if their id is in this list even if `add_as_admin` is false.
+    /// - `members` - List of users to be added as members of the group. This list takes priority over `add_as_member`,
+    ///             so the calling user will be added as a member if their id is in this list even if `add_as_member` is false.
     /// - `needs_rotation`
     ///     - true - group's private key will be marked for rotation
     ///     - false (default) - group's private key will not be marked for rotation
@@ -73,8 +91,8 @@ impl GroupCreateOpts {
         }
     }
 
-    fn standardize(self, calling_id: &UserId) -> GroupCreateOpts {
-        // if owner contains own id, set to None, as this is the server's default
+    fn standardize(self, calling_id: &UserId) -> GroupCreateOptsStd {
+        // if `owner` contains own id, set to None, as this is the server's default
         let owner = if self.owner == Some(calling_id.clone()) {
             None
         } else {
@@ -91,7 +109,7 @@ impl GroupCreateOpts {
         };
         let standardized_admins = {
             // if `add_as_admin`, make sure the calling user is in the `admins` list
-            let mut admins = if self.add_as_member && !self.admins.contains(calling_id) {
+            let mut admins = if self.add_as_admin && !self.admins.contains(calling_id) {
                 let mut admins = self.admins.clone();
                 admins.push(calling_id.clone());
                 admins
@@ -111,22 +129,28 @@ impl GroupCreateOpts {
             }
             admins
         };
-        GroupCreateOpts::new(
-            self.id,
-            self.name,
-            self.add_as_admin,
-            self.add_as_member,
-            owner,
-            standardized_admins,
-            standardized_members,
-            self.needs_rotation,
-        )
+
+        // concatenate the vectors of admin and member ids. duplicates will be removed later.
+        let all_users = [&standardized_admins[..], &standardized_members[..]].concat();
+        let set: HashSet<UserId> = HashSet::from_iter(all_users);
+        let users: Vec<UserId> = set.into_iter().collect();
+
+        GroupCreateOptsStd {
+            id: self.id,
+            name: self.name,
+            owner: owner,
+            admins: standardized_admins,
+            members: standardized_members,
+            users: users,
+            needs_rotation: self.needs_rotation,
+        }
     }
 }
 
 impl Default for GroupCreateOpts {
+    // Default GroupCreateOpts for common use cases. The user who calls `group_create()` will be the owner of the group
+    // as well as an admin and member of the group.
     fn default() -> Self {
-        // todo: docs
         GroupCreateOpts::new(None, None, true, true, None, vec![], vec![], false)
     }
 }
@@ -234,66 +258,25 @@ impl GroupOps for crate::IronOxide {
     fn group_create(&self, opts: &GroupCreateOpts) -> Result<GroupCreateResult> {
         let mut rt = Runtime::new().unwrap();
 
-        let GroupCreateOpts {
-            id: maybe_id,
-            name: maybe_name,
-            add_as_admin: _,
-            add_as_member: _,
+        let GroupCreateOptsStd {
+            id,
+            name,
             owner,
             admins,
             members,
+            users,
             needs_rotation,
         } = opts.clone().standardize(self.device.auth().account_id());
-
-        // if add_as_member {
-        //     members.push(self.device.auth().account_id().clone());
-        // };
-
-        // if add_as_admin {
-        //     admins.push(self.device.auth().account_id().clone());
-        // };
-
-        // concatenate the vectors of admin and member ids. duplicates will be removed later.
-        let users = [&admins[..], &members[..]].concat();
-        //need to add owner still
-
-        // let owner: Option<UserId>;
-        // match maybe_owner.clone() {
-        //     Some(id) => {
-        //         // the owner must be in the list of admins
-        //         if !admins.contains(&id) {
-        //             admins.push(id.clone());
-        //         }
-        //         if id == self.device.auth().account_id().clone() {
-        //             //Todo: test this
-        //             dbg!("Passed in self as owner");
-        //             owner = None;
-        //         } else {
-        //             users.push(id);
-        //             owner = maybe_owner;
-        //         }
-        //     }
-        //     None => {
-        //         if !admins.contains(self.device.auth().account_id()) {
-        //             admins.push(self.device.auth().account_id().clone());
-        //         }
-        //         owner = None;
-        //     }
-        // };
-
-        use std::{collections::HashSet, iter::FromIterator};
-        let set: HashSet<UserId> = HashSet::from_iter(users);
-        let users_to_lookup: Vec<UserId> = set.into_iter().collect();
 
         rt.block_on(group_api::group_create(
             &self.recrypt,
             self.device.auth(),
-            maybe_id,
-            maybe_name,
+            id,
+            name,
             owner,
-            &admins,
-            &members,
-            &users_to_lookup,
+            admins,
+            members,
+            &users,
             needs_rotation,
         ))
     }
--- a/src/group.rs
+++ b/src/group.rs
@@ -9,13 +9,16 @@ use crate::{
 use std::{collections::HashSet, iter::FromIterator};
 use tokio::runtime::current_thread::Runtime;
 
+// This is used for GroupCreateOpts that have been standardized with the GroupCreateOpts::standardize function.
+// `add_as_member` and `add_as_admin` have been removed, with the calling user added to the `members` and `admins` lists.
+// `all_users` holds all the users who need their public keys looked up with duplicates removed.
 struct GroupCreateOptsStd {
     id: Option<GroupId>,
     name: Option<GroupName>,
     owner: Option<UserId>,
     admins: Vec<UserId>,
     members: Vec<UserId>,
-    users: Vec<UserId>,
+    all_users: Vec<UserId>,
     needs_rotation: bool,
 }
 
@@ -92,13 +95,6 @@ impl GroupCreateOpts {
     }
 
     fn standardize(self, calling_id: &UserId) -> GroupCreateOptsStd {
-        // if `owner` contains own id, set to None, as this is the server's default
-        let owner = if self.owner == Some(calling_id.clone()) {
-            None
-        } else {
-            self.owner
-        };
-
         // if `add_as_member`, make sure the calling user is in the `members` list
         let standardized_members = if self.add_as_member && !self.members.contains(calling_id) {
             let mut members = self.members.clone();
@@ -116,7 +112,7 @@ impl GroupCreateOpts {
             } else {
                 self.admins
             };
-            match owner.clone() {
+            match self.owner.clone() {
                 Some(owner_id) => {
                     // if the owner is specified, make sure they're in the `admins` list
                     if !admins.contains(&owner_id) {
@@ -130,10 +126,18 @@ impl GroupCreateOpts {
             admins
         };
 
+        // if `owner` contains own id, set to None, as this is the server's default
+        let owner = if self.owner == Some(calling_id.clone()) {
+            None
+        } else {
+            self.owner
+        };
+
         // concatenate the vectors of admin and member ids. duplicates will be removed later.
-        let all_users = [&standardized_admins[..], &standardized_members[..]].concat();
-        let set: HashSet<UserId> = HashSet::from_iter(all_users);
-        let users: Vec<UserId> = set.into_iter().collect();
+        // owner should be in the list already, otherwise this will get caught later
+        let admins_and_members = [&standardized_admins[..], &standardized_members[..]].concat();
+        let set: HashSet<UserId> = HashSet::from_iter(admins_and_members);
+        let all_users: Vec<UserId> = set.into_iter().collect();
 
         GroupCreateOptsStd {
             id: self.id,
@@ -141,7 +145,7 @@ impl GroupCreateOpts {
             owner: owner,
             admins: standardized_admins,
             members: standardized_members,
-            users: users,
+            all_users: all_users,
             needs_rotation: self.needs_rotation,
         }
     }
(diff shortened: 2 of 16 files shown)


src/group.rs
```diff
@@ -6,19 +6,43 @@ use crate::{
...
+    // true (default) - creating user will be added as an admin of the group.
+    // false - creating user will not be added as an admin of the group.
+    add_as_admin: bool,
     // true (default) - creating user will be added to the group's membership (in addition to being the group's admin);
     // false - creating user will not be added to the group's membership
     add_as_member: bool,
-    // list of users to add as members to the group.
+    // None (default) - The creating user will be the owner of the group.
+    // Some(UserId) - The provided user will be the owner of the group.
+    // Note that the owner must be included in the `admins` list.
```
ernieturner: Should we maybe just take the user ID that was provided here and manually add it to the list of admins so that they don't have to specify it in two places?

giarc3: So that's actually what we are doing, I just have it documented weirdly. We decided on a semi-lenient way of requiring that the owner is in the admin list. If:
- `owner` is Some(UserId), I put the UserId in the admins list for them because they were explicit about choosing an owner.
- `owner` is None (which makes the calling user the owner), then they're hopefully using `add_as_admin` as well. If not, they did something wrong and are being confusing about what they want, so it will error on them for not being in the admins list. 

So we're kind of in a half-strict, half-lenient state where if they're explicit about the owner, we do it for them, but if they just leave it at None, they have to do it themselves (or with `add_as_admin`).
I'm open to suggestions about changing the documentation or implementation, this is just what we've decided on for now.

ernieturner: Ok, I think the logic is fine, I'd just change this wording to something like

```
    // Specifies who the owner of this group is. Group owners have the same permissions as other admins but they cannot be removed as admins from this group.
    // None (default) - The creating user will be the owner of the group. Cannot be used if `add_as_admin` is set to false as the group must have an owner.
    // Some(UserId) - The provided user will be the owner of the group. This ID will automatically be added to the admins list.
```

giarc3: Thanks for the feedback. I've added in this distinction to make it more clear.

src/internal/group_api/mod.rs
```diff
@@ -140,12 +140,14 @@ impl GroupMetaResult {
     }
 }
 
+#[derive(Debug)]
```
clintfred: Going forward we want all *Result types to be Debug, Clone, PartialEq. See #58 
```suggestion
#[derive(Debug, Clone, PartialEq)]
```

src/internal/group_api/mod.rs
```diff
@@ -359,88 +372,153 @@ pub(crate) fn get_group_keys<'a>(
...
     recrypt: &'a Recrypt<Sha256, Ed25519, RandomBytes<CR>>,
     auth: &'a RequestAuth,
-    user_master_pub_key: &'a PublicKey,
     group_id: Option<GroupId>,
     name: Option<GroupName>,
-    members: &'a Vec<UserId>,
+    owner: Option<UserId>,
+    admins: Vec<UserId>,
+    members: Vec<UserId>,
+    users_to_lookup: &'a Vec<UserId>,
```
clintfred: I think this should be computed dynamically in this function, or if you add `admin_and_members` to `GroupCreateOpsStd`, perhaps you should just pass the `GroupCreateOpsStd` struct into the internal api.

src/internal/group_api/mod.rs
```diff
@@ -359,88 +372,153 @@ pub(crate) fn get_group_keys<'a>(
...
-                                maybe_transform_key.map(|transform_key| {
-                                    (id, (user_pub_key, Some(transform_key.into())))
-                                })
+    use futures::future::Either;
+    let owner_id = match owner.clone() {
+        Some(id) => id,
+        None => auth.account_id().clone(),
+    };
+    // don't make any requests if validation fails
+    if admins.is_empty() {
```
clintfred: Maybe this validation should be moved into `GroupCreateOpsStd`? Be nice to not even get here if the input isn't valid.

src/internal/group_api/mod.rs
```diff
@@ -359,88 +372,153 @@ pub(crate) fn get_group_keys<'a>(
...
+            format!("admins"),
+            format!("admins list cannot be empty"),
+        )))
+    } else if !admins.contains(&owner_id) {
+        Either::A(err(IronOxideErr::ValidationError(
+            format!("admins"),
+            format!("admins list must contain the owner"),
+        )))
+    } else {
+        Either::B(
```
clintfred: I think we need to figure out a way to break this branch up. Can we make some smaller functions here that could be tested?

src/internal/group_api/mod.rs
```diff
@@ -359,88 +372,153 @@ pub(crate) fn get_group_keys<'a>(
...
+            user_api::user_key_list(auth, users_to_lookup)
+                .and_then(move |user_ids_and_keys| {
+                    // this will occur when one of the UserIds cannot be found
+                    if user_ids_and_keys.len() != users_to_lookup.len() {
+                        // figure out which user ids could not be found in the database and include them in the error message.
+                        let desired_users_set: HashSet<UserId> =
```
clintfred: Would it be better to sort the two vectors and then just check equality? I think Vec probably has a PartialEq to do this...

src/internal/group_api/mod.rs
```diff
@@ -359,88 +372,153 @@ pub(crate) fn get_group_keys<'a>(
...
+                            diff
+                        )))
+                    } else {
+                        transform::gen_group_keys(recrypt)
+                            .and_then(move |(plaintext, group_priv_key, group_pub_key)| {
+                                let (member_info_result, admin_info): (
+                                    Result<Vec<(UserId, PublicKey, TransformKey)>, IronOxideErr>,
+                                    Vec<(UserId, PublicKey)>,
+                                ) = {
+                                    let members_set: HashSet<_> = HashSet::from_iter(&members);
```
clintfred: Why are we making these HashSets? To remove dupes?

src/internal/group_api/mod.rs
```diff
@@ -359,88 +372,153 @@ pub(crate) fn get_group_keys<'a>(
...
+                                    let admins_set: HashSet<_> = HashSet::from_iter(&admins);
+                                    let mut member_info: Vec<
+                                        Result<(UserId, PublicKey, TransformKey), IronOxideErr>,
+                                    > = vec![];
+                                    let mut admin_info: Vec<(UserId, PublicKey)> = vec![];
+                                    user_ids_and_keys
```
clintfred: Would it be easier to put the ids back into two lists earlier: members and admins -- to simplify this? 
Step 1: partition the user_ids_and_keys
Step 2: Generate transform keys for members

src/internal/group_api/requests.rs
```diff
@@ -207,71 +211,46 @@ pub mod group_list {
...
-                };
-
-                auth.request.post(
-                    "groups",
-                    &req,
-                    RequestErrorCode::GroupCreate,
-                    AuthV2Builder::new(&auth, Utc::now()),
-                )
-            })
+        let req = GroupCreateReq {
```
clintfred: Really nice clean-up here.

src/internal/group_api/mod.rs
```diff
@@ -359,88 +372,153 @@ pub(crate) fn get_group_keys<'a>(
...
+        Some(id) => id,
+        None => auth.account_id().clone(),
+    };
+    // don't make any requests if validation fails
+    if admins.is_empty() {
+        Either::A(err(IronOxideErr::ValidationError(
+            format!("admins"),
+            format!("admins list cannot be empty"),
+        )))
+    } else if !admins.contains(&owner_id) {
```
ToniaSchlick: You can just check if add_as_admin is false and owner = none

src/group.rs
```diff
@@ -6,19 +6,43 @@ use crate::{
     internal::{group_api, user_api::UserId},
     Result,
 };
+use std::{collections::HashSet, iter::FromIterator};
 use tokio::runtime::current_thread::Runtime;
 
+// This is used for GroupCreateOpts that have been standardized with the GroupCreateOpts::standardize function.
+// `add_as_member` and `add_as_admin` have been removed, with the calling user added to the `members` and `admins` lists.
+// `all_users` holds all the users who need their public keys looked up with duplicates removed.
+struct GroupCreateOptsStd {
```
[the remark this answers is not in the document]
giarc3: I am not well informed yet on what goes where, so I'll take your word for it and move this over.

src/group.rs
```diff
@@ -6,19 +6,43 @@ use crate::{
...
     // human readable name of the group. Does not need to be unique.
     name: Option<GroupName>,
+    // true (default) - creating user will be added as an admin of the group.
+    // false - creating user will not be added as an admin of the group.
+    add_as_admin: bool,
     // true (default) - creating user will be added to the group's membership (in addition to being the group's admin);
     // false - creating user will not be added to the group's membership
     add_as_member: bool,
-    // list of users to add as members to the group.
+    // Specifies who the owner of this group is. Group owners have the same permissions as other admins but they cannot be removed as admins from this group.
```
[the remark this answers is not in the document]
giarc3: Changed.

src/group.rs
```diff
@@ -32,52 +56,106 @@ impl GroupCreateOpts {
...
+        let standardized_admins = {
+            // if `add_as_admin`, make sure the calling user is in the `admins` list
+            let mut admins = if self.add_as_admin && !self.admins.contains(calling_id) {
+                let mut admins = self.admins.clone();
+                admins.push(calling_id.clone());
+                admins
+            } else {
+                self.admins
+            };
+            match self.owner.clone() {
```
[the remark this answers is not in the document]
giarc3: I think I did this. Please check it

src/internal/group_api/mod.rs
giarc3: This would miss the case where the caller adds themselves to the `admins` list manually, which we decided takes precedence over `add_as_admin`. But I realized that my `if` and `else if` really only covered one thing together, so I cleaned them up as a single check.

src/group.rs
```diff
@@ -32,52 +56,106 @@ impl GroupCreateOpts {
...
+                    }
+                }
+                // if the owner is the calling user (default), they should have been added to the
+                // admins list by `add_as_admin`. If they aren't it will error later on.
+                None => (),
+            }
+            admins
+        };
+
+        // if `owner` contains own id, set to None, as this is the server's default
```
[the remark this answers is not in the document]
coltfred: You can do it either way. There's no reason to clear it if they sent it IMO and then you can get rid of this code.

src/internal/group_api/mod.rs
```diff
@@ -359,88 +372,153 @@ pub(crate) fn get_group_keys<'a>(
...
+                    if user_ids_and_keys.len() != users_to_lookup.len() {
+                        // figure out which user ids could not be found in the database and include them in the error message.
+                        let desired_users_set: HashSet<UserId> =
+                            HashSet::from_iter(users_to_lookup.clone());
+                        let found_users_set: HashSet<UserId> =
+                            user_ids_and_keys.into_iter().map(|(x, _)| x).collect();
+                        let diff: HashSet<&UserId> =
+                            desired_users_set.difference(&found_users_set).collect();
+                        err(IronOxideErr::UserDoesNotExist(format!(
+                            "Failed to find the following users: {:?}",
```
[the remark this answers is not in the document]
coltfred: `{}` vs `{:?}` One prints the debug version of the struct and one prints the display version.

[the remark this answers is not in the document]
clintfred: `{}` is Display whereas `{:?}` is Debug. We probably don't ever want to use Debug in error messages.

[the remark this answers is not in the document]
giarc3: Using `{}` fails to compile: ````std::collections::HashSet<&internal::user_api::UserId>` doesn't implement `std::fmt::Display````.

Using `{:?}` gives the IronOxideErr: `Error: UserDoesNotExist("Failed to find the following users: {UserId(\"aretheseuserids\"), UserId(\"likeafox\"), UserId(\"whatsthenextwhat\")}")`.

Is that not what we'd want it to display? If so, how do I implement Display for a HashSet of &UserId?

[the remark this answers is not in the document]
clintfred: I don't think any of the `std::collections` implement Display. This might be helpful https://doc.rust-lang.org/rust-by-example/hello/print/print_display.html

src/group.rs
```diff
@@ -6,19 +6,43 @@ use crate::{
...
+// This is used for GroupCreateOpts that have been standardized with the GroupCreateOpts::standardize function.
+// `add_as_member` and `add_as_admin` have been removed, with the calling user added to the `members` and `admins` lists.
+// `all_users` holds all the users who need their public keys looked up with duplicates removed.
+struct GroupCreateOptsStd {
+    id: Option<GroupId>,
+    name: Option<GroupName>,
+    owner: Option<UserId>,
+    admins: Vec<UserId>,
+    members: Vec<UserId>,
+    all_users: Vec<UserId>,
```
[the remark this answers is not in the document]
clintfred: `[UserId]` is an array (stack allocated) opposed to `Vec` which is heap allocated. In this case `Vec` is probably what you want as you need to allocate the memory to hold your concat'd list. 

Feel free to resolve this unless you have other questions.

src/group.rs
```diff
@@ -6,19 +6,43 @@ use crate::{
...
 #[derive(Clone)]
 /// Options for group creation.
 pub struct GroupCreateOpts {
     // unique id of a group within a segment. If none, the server will assign an id.
     id: Option<GroupId>,
     // human readable name of the group. Does not need to be unique.
     name: Option<GroupName>,
+    // true (default) - creating user will be added as an admin of the group.
+    // false - creating user will not be added as an admin of the group.
+    add_as_admin: bool,
```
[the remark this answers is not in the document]
clintfred: I agree on the consistency point. Not sure if it's worth it...

src/group.rs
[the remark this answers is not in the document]
clintfred: I think depending on a server default is less good as it leaves room for a bug to be introduced if the server's default changes.

src/group.rs
```diff
@@ -32,52 +56,106 @@ impl GroupCreateOpts {
...
+        // if `owner` contains own id, set to None, as this is the server's default
+        let owner = if self.owner == Some(calling_id.clone()) {
+            None
+        } else {
+            self.owner
+        };
+
+        // concatenate the vectors of admin and member ids. duplicates will be removed later.
+        // owner should be in the list already, otherwise this will get caught later
+        let admins_and_members = [&standardized_admins[..], &standardized_members[..]].concat();
```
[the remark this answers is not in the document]
clintfred: chain doesn't allocate. In this case I think you need to allocate, so concat is fine.

src/internal/group_api/mod.rs
clintfred: Fair enough.

src/internal/group_api/mod.rs
clintfred: I wasn't meaning literal `partition`, but the idea of separating the members and admins earlier. I still think this might be a good idea, but I'll look at the latest code and see...

src/group.rs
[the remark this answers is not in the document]
giarc3: Would you propose that I instead never use the server default? I could change my `GroupCreateOptsStd.owner` to be just a `UserId` (not an option), and always send it in. Then this chunk of code would go the other direction match on the option to fill `owner` either with the Some's value or the calling user's id.

src/internal/group_api/mod.rs
```diff
@@ -359,88 +372,153 @@ pub(crate) fn get_group_keys<'a>(
...
+            format!("admins"),
+            format!("admins list must contain the owner"),
+        )))
+    } else {
+        Either::B(
+            user_api::user_key_list(auth, users_to_lookup)
+                .and_then(move |user_ids_and_keys| {
+                    // this will occur when one of the UserIds cannot be found
+                    if user_ids_and_keys.len() != users_to_lookup.len() {
+                        // figure out which user ids could not be found in the database and include them in the error message.
```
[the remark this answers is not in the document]
giarc3: Refactored into a generic `compare_users()` function and added a unit test.

src/internal/group_api/mod.rs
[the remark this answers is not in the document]
giarc3: I factored this out into `collect_admin_and_member_info()`, but it takes so many different things (because it generates transform keys for each member) that I'm having trouble including a test for this specifically.

src/group.rs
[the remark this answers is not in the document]
giarc3: I removed this section of code altogether, so it now just keeps `owner` as the option from `GroupCreateOpts`, and the server will respond the same whether it's `None` or `Some(calling_user)`

src/group.rs
```diff
@@ -32,52 +56,106 @@ impl GroupCreateOpts {
...
             add_as_member,
+            owner,
+            admins,
             members,
             needs_rotation,
         }
     }
 
-    fn standardize(self, calling_id: &UserId) -> GroupCreateOpts {
+    fn standardize(self, calling_id: &UserId) -> GroupCreateOptsStd {
```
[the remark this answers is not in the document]
giarc3: I added several unit tests for going from `GroupCreateOpts` to `GroupCreateOptsStd`

src/internal/group_api/mod.rs
giarc3: The code has been broken down into some smaller functions. Let me know if you think it should be broken down further.

src/group.rs
```diff
@@ -32,52 +42,100 @@ impl GroupCreateOpts {
...
             add_as_member,
+            owner,
+            admins,
             members,
             needs_rotation,
         }
     }
 
-    fn standardize(self, calling_id: &UserId) -> GroupCreateOpts {
+    pub(crate) fn standardize(self, calling_id: &UserId) -> Result<GroupCreateOptsStd> {
```
clintfred: Why is this pub(crate)?

src/group.rs
```diff
@@ -32,52 +42,100 @@ impl GroupCreateOpts {
...
+            let owner: &UserId = match &self.owner {
+                Some(owner_id) => {
+                    // if the owner is specified, make sure they're in the `admins` list
+                    if !admins.contains(owner_id) {
+                        admins.push(owner_id.clone());
+                    }
+                    owner_id
+                }
+                // if the owner is the calling user (default), they should have been added to the
+                // admins list by `add_as_admin`. If they aren't, it will error later on.
```
clintfred: what does this mean "will error later on"?

src/group.rs
```diff
@@ -32,52 +42,100 @@ impl GroupCreateOpts {
...
+                admins: standardized_admins,
+                members: standardized_members,
+                needs_rotation: self.needs_rotation,
+            })
+        }
     }
 }
 
 impl Default for GroupCreateOpts {
+    // Default GroupCreateOpts for common use cases. The user who calls `group_create()` will be the owner of the group
```
clintfred: If these have `///` they _might_ show up on the doc site.

src/internal/group_api/mod.rs
```diff
@@ -2,30 +2,50 @@ use crate::{
...
 };
-
 mod requests;
 
 pub enum GroupEntity {
     Member,
     Admin,
 }
 
+// This is used for GroupCreateOpts that have been standardized with the GroupCreateOpts::standardize function.
```
clintfred: `///`

src/internal/group_api/mod.rs
```diff
@@ -351,6 +384,63 @@ pub(crate) fn get_group_keys<'a>(
...
 
+fn compare_users<T: Eq + std::hash::Hash + std::fmt::Debug, X>(
+    desired_users: &Vec<T>,
+    found_users: HashMap<T, X>,
+) -> IronOxideErr {
+    let desired_users_set: HashSet<&T> = HashSet::from_iter(desired_users);
+    let found_users_vec: Vec<T> = found_users.into_iter().map(|(x, _)| x).collect();
+    let found_users_set: HashSet<&T> = HashSet::from_iter(&found_users_vec);
+    let diff: Vec<&&T> = desired_users_set.difference(&found_users_set).collect();
+    IronOxideErr::UserDoesNotExist(format!("Failed to find the following users: {:?}", diff))
```
clintfred: Still using Debug here.

src/internal/group_api/mod.rs
```diff
@@ -351,6 +384,63 @@ pub(crate) fn get_group_keys<'a>(
...
+                    &user_pub_key.clone().into(),
+                    &signing_key.into(),
+                );
+                match maybe_transform_key {
+                    Ok(member_trans_key) => member_info.push(Ok((
+                        id.clone(),
+                        user_pub_key.clone(),
+                        member_trans_key.into(),
+                    ))),
+                    Err(_) => (),
```
clintfred: We should probably do something with this error case.

src/internal/group_api/mod.rs
```diff
@@ -351,6 +384,63 @@ pub(crate) fn get_group_keys<'a>(
...
+fn collect_admin_and_member_info<CR: rand::CryptoRng + rand::RngCore>(
+    recrypt: &Recrypt<Sha256, Ed25519, RandomBytes<CR>>,
+    signing_key: &crate::internal::DeviceSigningKeyPair,
+    group_priv_key: recrypt::api::PrivateKey,
+    admins: Vec<UserId>,
+    members: Vec<UserId>,
+    user_ids_and_keys: HashMap<UserId, PublicKey>,
+) -> Result<
+    (
+        Vec<(UserId, PublicKey, TransformKey)>,
```
clintfred: I think generating the transform keys in this function is not good separation of concerns. You could argue that if the number of members is huge there could be a performance reason to mix this in here so that you don't have to go over the list again, but I don't see that as a legitimate concern.

src/internal/group_api/mod.rs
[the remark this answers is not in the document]
giarc3: Okay, I was able to make tests on this function, but I'm not sure what good test cases would be. I just have a test with filled admins and members lists, and a test with an empty members list. The admins list will have been validated to never be empty ahead of time.

src/internal/group_api/mod.rs
clintfred: I think this will get easier with the switch to `async fn` as impl Future seems more temperamental...

src/internal/group_api/mod.rs
```diff
@@ -360,90 +450,94 @@ pub(crate) fn get_group_keys<'a>(
...
+                                        encrypted_msg: enc_msg,
+                                        user: requests::User {
+                                            user_id: admin_id,
+                                            user_master_public_key: admin_pub_key.into(),
+                                        },
+                                    })
                             })
                             .collect();
-                        let member_info = member_info_result?;
+                        let group_admins = group_admins_result?;
```
clintfred: Can this be combined onto the previous statement?

src/internal/group_api/mod.rs
clintfred: You implemented what I was thinking, except for the combining of the transform key generation.

src/internal/group_api/mod.rs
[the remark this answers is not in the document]
clintfred: Consider using https://github.com/dac-gmbh/vec1 for NonEmptyList and get rid of error cases at compile time :)

We want to bring this in anyway...
https://github.com/IronCoreLabs/ironoxide/issues/18#issuecomment-519352993

src/group.rs
giarc3: The validation for it is a few lines below it. I was trying to show that the case is accounted for, even though it's not right here. Should I call this out differently?

src/internal/group_api/mod.rs
giarc3: I still haven't found an alternate way to display it. I tried implementing Display for UserId, but it didn't seem to make a difference (unless I did it wrong).
The output from Debug looks reasonable to me, so is there a concern with leaving it?

src/group.rs
giarc3: I don't remember. I added it at some point when it said the function was private and I needed it, but I assume moving things around fixed that? So I've removed it now

src/group.rs
giarc3: Changed

src/internal/group_api/mod.rs
giarc3: Changed

src/internal/group_api/mod.rs
giarc3: I see this function as getting everything ready to make `request::GroupMember`s and `request::GroupAdmin`s, so generating the `TransformKey`s here seems to fulfill that purpose to me. Are you thinking of it more as just a way to re-partition the admins and members?

src/internal/group_api/mod.rs
giarc3: Good catch. I had meant to push that error into the `member_info` vector so that the `.collect()` at the end would successfully fail on the `?`. I did that now.

src/internal/group_api/mod.rs
giarc3: I'm hoping so

src/internal/group_api/mod.rs
giarc3: It can't (as far as I've tried) because the `.collect()` uses the type I've told it to know what to collect it into, which would then be a different type after the `?`. So it seems to get confused if I combine them.

giarc3: Okay, never mind. In a completely unrelated comment about turbofish, @skeet70 made me realize that I can do this in one line. So I did that.

src/internal/group_api/mod.rs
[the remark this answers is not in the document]
giarc3: I made `admins` a NonEmptyList in `GroupCreateOptsStd` and throughout the `group_create` calls

src/internal/group_api/mod.rs
coltfred: I think debug is fine for this case. The error looks good, which is what we want.

clintfred: I did get this to work by making 
```rust
struct DisplayableVec<T: Display>(Vec<T>);
```

I think I'm going to back off this. The error itself is Display, so that's good. We have other places where we are using Debug to build up the error. If we want to address this, we address this as well.

src/internal/group_api/mod.rs
coltfred: Maybe we should rename the function to note that, or add a comment to it? Really it wouldn't be _that_ bad to have this return the 2 vecs and feed the members vec back into another function which would generate the transform keys. At the same time, this does break it up more than it was, which is incremental improvement.

I'll let you and clint figure out what needs to be done. :running_man:

clintfred: I still do think it would be cleaner to do the transform key generation as a separate step, but this solution isn't bad. I think the function signature makes it clear that transform keys are also being computed in here, but maybe putting a small doc comment on this function is a good idea anyway.

CHANGELOG.md
```diff
@@ -2,8 +2,11 @@
 
 ## 0.14.0 (Unreleased)
 
+- [[#76](https://github.com/IronCoreLabs/ironoxide/pull/76)]
+  - Allows adding admins at group creation time.
+  - Allow specifying an owner at group creation time.
```
clintfred:
```suggestion
  - Allows specifying an owner at group creation time.
```

giarc3: Fixed

Issue #73, opened 2019-11-19
Update internal interfaces to use async fn


clintfred:
As a ironoxide developer,
I want the internal APIs to use `async` fn instead of `Future` (0.1)
so that the code will be easier to maintain and integrate with other async Rust libraries

AC: 
* [x] User API
* [x] user/requests
* [ ] Group API
* [ ] group/requests
* [ ] Document API
* [ ] document/requests
* [ ] rest.rs

Not included in this issue:
* creating a public async API
* trying to cache the tokio runtime
* trying to cache reqwest clients

PR #77, opened 2019-11-20
[#73] Converts internal user api and user/requests to async fn

clintfred:
see #73 

Functions in the internal user API and request layer have been updated to use `async fn` instead of `impl Future` or `Box<dyn Future>`. Places outside of the user API that used user functions were not updated, but rather the new Futures were converted back to Futures 0.1. 

Note that `internal/user_api/mod.rs` and `internal/user_api/requests.rs` no longer import anything from the futures (0.1) crate.

The intent is to do the group_api, document_api, rest.rs in separate PRs.

### DIFF
--- a/src/internal/user_api/mod.rs
+++ b/src/internal/user_api/mod.rs
@@ -330,69 +330,61 @@ pub fn user_rotate_private_key<'apicall, CR: rand::CryptoRng + rand::RngCore>(
 }
 
 /// Generate a device key for the user specified in the JWT.
-pub fn generate_device_key<'a, CR: rand::CryptoRng + rand::RngCore>(
+pub async fn generate_device_key<'a, CR: rand::CryptoRng + rand::RngCore>(
     recrypt: &'a Recrypt<Sha256, Ed25519, RandomBytes<CR>>,
     jwt: &'a Jwt,
     password: Password,
     device_name: Option<DeviceName>,
     signing_ts: &'a DateTime<Utc>,
     request: &'a IronCoreRequest<'static>,
-) -> impl Future<Item = DeviceContext, Error = IronOxideErr> + 'a {
+) -> Result<DeviceContext, IronOxideErr> {
     // verify that this user exists
-    requests::user_verify::user_verify(&jwt, &request)
-        .boxed()
-        .compat()
+    let requests::user_verify::UserVerifyResponse {
+        user_private_key,
+        user_master_public_key,
+        id: account_id,
+        segment_id,
+        ..
+    } = requests::user_verify::user_verify(&jwt, &request)
+        .await
         .and_then(|maybe_user| {
             maybe_user.ok_or(IronOxideErr::UserDoesNotExist(
                 "Device cannot be added to a user that doesn't exist".to_string(),
             ))
-        })
-        // unpack the verified user and create a DeviceAdd
-        .and_then(
-            move |requests::user_verify::UserVerifyResponse {
-                      user_private_key,
-                      user_master_public_key,
-                      id: account_id,
-                      segment_id,
-                      ..
-                  }| {
-                Ok((
-                    {
-                        let user_public_key: RecryptPublicKey =
-                            PublicKey::try_from(user_master_public_key)?.into();
-                        let user_private_key =
-                            EncryptedMasterKey::new_from_slice(&user_private_key.0)?;
+        })?;
+    // unpack the verified user and create a DeviceAdd
+    let (device_add, account_id, segment_id) = (
+        {
+            let user_public_key: RecryptPublicKey =
+                PublicKey::try_from(user_master_public_key)?.into();
+            let user_private_key = EncryptedMasterKey::new_from_slice(&user_private_key.0)?;
 
-                        // decrypt the user's master key using the provided password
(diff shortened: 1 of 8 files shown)


src/internal/user_api/mod.rs
```diff
@@ -265,195 +267,189 @@ impl UserUpdatePrivateKeyResult {
...
+    };
+    requests::user_update_private_key::update_private_key(
+        &auth,
+        user_id,
+        curr_key_id,
+        new_encrypted_priv_key.into(),
+        aug_factor.into(),
+    )
+    .await
+    .map(|resp| resp.into())
```
coltfred: Seems like you should be able to `?.into()` now.

src/internal/user_api/mod.rs
```diff
@@ -265,195 +267,189 @@ impl UserUpdatePrivateKeyResult {
...
     // verify that this user exists
-    requests::user_verify::user_verify(&jwt, &request)
+    let requests::user_verify::UserVerifyResponse {
+        user_private_key,
+        user_master_public_key,
+        id: account_id,
+        segment_id,
+        ..
+    } = requests::user_verify::user_verify(&jwt, &request)
+        .await
```
coltfred: Can't you do `?` on here to avoid the `and_then` nesting?

src/internal/user_api/mod.rs
```diff
@@ -265,195 +267,189 @@ impl UserUpdatePrivateKeyResult {
...
+            let device_add = generate_device_add(recrypt, &jwt, &user_keypair, &signing_ts)?;
+            device_add
+        },
+        account_id.try_into()?,
+        segment_id,
+    );
+
+    // call device_add
+    requests::device_add::user_device_add(&jwt, &device_add, &device_name, &request)
+        .await
```
coltfred: Can you do `?` here and just do `Ok` on the expression in the move?

src/internal/user_api/mod.rs
```diff
@@ -265,195 +267,189 @@ impl UserUpdatePrivateKeyResult {
...
         })
 }
 
-pub fn device_list(
-    auth: &RequestAuth,
-) -> impl Future<Item = UserDeviceListResult, Error = IronOxideErr> + '_ {
-    requests::device_list::device_list(auth).map(|resp| {
+pub async fn device_list(auth: &RequestAuth) -> Result<UserDeviceListResult, IronOxideErr> {
+    let resp = requests::device_list::device_list(auth).await?;
+    {
```
coltfred: You left the { because of the mut?

src/lib.rs
```diff
@@ -148,32 +148,37 @@ impl PrivateKeyRotationCheckResult {
...
-                        rotations_needed: EitherOrBoth::Left(curr_user.account_id().clone()),
-                    },
-                ))
-            } else {
-                Ok(InitAndRotationCheck::NoRotationNeeded(ironoxide))
-            }
-        }),
+        internal::user_api::user_get_current(device_context.auth())
+            .boxed()
+            .compat()
```
coltfred: Can you await to avoid the `and_then`?

src/user.rs
```diff
@@ -172,38 +181,58 @@ impl UserOps for IronOxide {
...
     }
 
     fn user_rotate_private_key(&self, password: &str) -> Result<UserUpdatePrivateKeyResult> {
         let mut rt = Runtime::new().unwrap();
-        rt.block_on(user_api::user_rotate_private_key(
-            &self.recrypt,
-            password.try_into()?,
-            self.device().auth(),
-        ))
+        rt.block_on(
```
coltfred: Did you consider going to the new tokio runtime and bringing all the 0.1 futures into 0.3 to run them instead of doing all this work (which we'll immediately have to undo).

src/user.rs
```diff
@@ -172,38 +181,58 @@ impl UserOps for IronOxide {
...
+            user_api::user_rotate_private_key(
+                &self.recrypt,
+                password.try_into()?,
+                self.device().auth(),
+            )
+            .boxed_local()
```
coltfred: Why was boxed_local needed here instead of `boxed`?

src/user.rs
clintfred: I did consider it, but this seemed more incremental (and worked!). I think I would need the unlreased `tokio-compat` to run v1 futures on new Tokio, possibly. This way the tokio work can be in its own PR (and ticket, probably).

src/internal/user_api/mod.rs
clintfred: Have to also wrap the whole think in an `Ok`, but maybe that's still better?

src/internal/user_api/mod.rs
clintfred: reworked slightly....

src/lib.rs
clintfred: Can't await here unless we make this function async or use an async block. This is probably best handled when we upgrade tokio and starting caching the runtime.

src/user.rs
clintfred: I haven't fully understood why we have to do that, yet. There are a couple of others. I added a comment here.

PR #81, opened 2019-11-26
Move to tokio 0.2 and futures 3 and make groups api async

coltfred:
- Remove futures 0.1 dependency
- Upgrade to tokio 0.2
- Change group apis to be async/await
- Temp fix for over encoding in reqwest for signing

TODO
- [x] Fix test
- [x] Figure out if we can be ok with url and signed url being different.


Cargo.toml
```diff
@@ -17,21 +17,19 @@ base64-serde = "~0.3.1"
 bytes = "~0.4.12"
 serde = "~1.0"
 serde_json = "~1.0"
-publicsuffix = "=1.5.2" # exact version needed to match url used by reqwest 0.9
+publicsuffix = "^1.5.4" # exact version needed to match url used by reqwest 0.9
```
clintfred: Remove comment about exact version? Should we use `~1.5.4` here?

src/internal/group_api/mod.rs
```diff
@@ -327,62 +326,56 @@ impl GroupAccessEditResult {
...
+    }?;
 
-    resp.and_then(|GroupListResponse { result }| {
-        let group_list = result
-            .into_iter()
-            .map(|g| g.try_into())
-            .collect::<Result<Vec<_>, _>>()?;
-        Ok(GroupListResult::new(group_list))
-    })
+    let GroupListResponse { result } = resp;
```
clintfred: You could merge this with `let resp` on line 333.

src/internal/group_api/mod.rs
```diff
@@ -327,62 +326,56 @@ impl GroupAccessEditResult {
...
-                    match maybe_public_key {
-                        Some(public_key) => Either::Right(WithKey {
-                            id: group_id,
-                            public_key,
-                        }),
-                        None => Either::Left(group_id),
-                    }
-                })
+    let list_resp = list(auth, Some(&groups)).await;
+    let GroupListResult { result } = list_resp?;
```
clintfred: Could merge with `let list_resp =` above.

src/internal/rest.rs
```diff
@@ -1,21 +1,17 @@
...
     auth_v2::AuthV2Builder, user_api::UserId, DeviceSigningKeyPair, IronOxideErr, Jwt,
     RequestErrorCode, OUR_REQUEST,
 };
-use futures::future::Either;
+use bytes::Bytes;
+use chrono::{DateTime, Utc};
 use percent_encoding::SIMPLE_ENCODE_SET;
+use reqwest::{
+    header::HeaderMap,
+    r#async::{Client as RClient, Request as ARequest},
```
skeet70:
```suggestion
    r#async::{Client as RClient, Request},
```

Can this just be `Request` now, or something more specific like `Reqwest` if the intent is to differentiate it from `IronCoreRequest`?

skeet70: @clintfred ^

src/internal/rest.rs
```diff
@@ -358,77 +358,73 @@ impl<'a> IronCoreRequest<'a> {
...
+            let mut req = ARequest::new(
+                Method::POST,
+                format!("{}{}", self.base_url(), relative_url)
+                    .into_url()
+                    .map_err(|e| IronOxideErr::from((e, error_code)))?,
+            );
+            *req.body_mut() = Some(body.to_vec().into());
+            (req, body.to_vec())
+        });
+        let (mut req, body_bytes) = make_req?; //COLT: See if you can do better
```
skeet70: Is this comment to yourself for later or have you already tried to do better and forgot to remove it?

src/internal/rest.rs
clintfred: Pretty sure the `async` module is deprecated... maybe? Also there are two reqwest use blocks here. They should probably be combined.

src/internal/rest.rs
```diff
@@ -185,9 +184,9 @@ impl<'a> Authorization<'a> {
 pub struct SignatureUrlString(String);
 
 impl SignatureUrlString {
-    pub fn new(encoded_full_url: &str) -> Result<SignatureUrlString, reqwest::UrlError> {
+    pub fn new(encoded_full_url: &str) -> Result<SignatureUrlString, url::ParseError> {
```
clintfred: did reqwest stop re-exporting this type or did you think the `url` one was better?

src/internal/rest.rs
```diff
@@ -185,9 +184,9 @@ impl<'a> Authorization<'a> {
...
         let parsed_url = Url::parse(encoded_full_url)?;
-        let query_str_format = |q: &str| format!("?{}", q);
+        let query_str_format = |q: &str| format!("?{}", q.replace("%27", "'"));
```
clintfred: There should probably be a comment here explaining what's going on here.

src/internal/rest.rs
clintfred: COLT comment

src/internal/rest.rs
```diff
@@ -1338,7 +1332,7 @@ mod tests {
             maybe_path.unwrap().signature_string(),
             "/api/1/users?id=user-10"
         );
-
+        //
```
skeet70:
```suggestion
```

src/internal/rest.rs
```diff
@@ -559,93 +560,87 @@ impl<'a> IronCoreRequest<'a> {
...
+            (req, body_bytes)
+        });
+        let (mut req, body_bytes) = make_req?;
+
+        // use the completed request to finish authorization v2 headers
+        let maybe_auth = SignatureUrlString::new(req.url().as_str())
+            .map(|sig_url| auth_b.finish_with(sig_url, req.method().clone(), Some(&body_bytes)))
+            .map_err(|e| IronOxideErr::from((e, error_code)));
+
+        // we only support Authorization::Version2 with this call
```
clintfred: There was some hope that the duplicated code between here and `post_raw` might be able to be easily refactored once we got rid of v1 futures. Is that worth taking a pass at, do you think?

src/internal/rest.rs
```diff
@@ -1371,6 +1365,7 @@ mod tests {
     }
 
     #[test]
+
```
skeet70:
```suggestion
```

src/internal/rest.rs
clintfred: remove?

src/internal/rest.rs
clintfred: I agree that we should just leave it as `Request`. `ARequest` actually came in when we were also using the blocking (then default) `Request` as well.

clintfred:
You should also write the change log entry for moving to futures 3, async/await, and tokio 0.2

src/internal/rest.rs
```diff
@@ -358,77 +358,73 @@ impl<'a> IronCoreRequest<'a> {
...
+                            replace_headers(req.headers_mut(), user_context_header);
+                            replace_headers(req.headers_mut(), DEFAULT_HEADERS.clone());
+                            replace_headers(req.headers_mut(), auth.to_auth_header());
+                            replace_headers(req.headers_mut(), request_sig.to_header());
+
+                            let result = Self::send_req(req, error_code, move |server_resp| {
+                                IronCoreRequest::deserialize_body(server_resp, error_code)
+                            })
+                            .await;
+                            result
```
giarc3: Unnecessary assignment

src/internal/group_api/mod.rs
```diff
@@ -463,98 +456,94 @@ pub fn group_create<'a, CR: rand::CryptoRng + rand::RngCore>(
...
-                member_vec,
-                needs_rotation,
+) -> Result<GroupCreateResult, IronOxideErr> {
+    let user_ids_and_keys = user_api::user_key_list(auth, users_to_lookup).await?;
+    // this will occur when one of the UserIds cannot be found
+    let (group_pub_key, member_vec, admin_vec) = if user_ids_and_keys.len() != users_to_lookup.len()
+    {
+        Err(compare_users(users_to_lookup, user_ids_and_keys))
+    } else {
+        let (plaintext, group_priv_key, group_pub_key) = transform::gen_group_keys(recrypt)?;
```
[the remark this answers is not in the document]
coltfred: I changed things... let me know if you like it more.

src/internal/group_api/mod.rs
```diff
@@ -463,98 +456,94 @@ pub fn group_create<'a, CR: rand::CryptoRng + rand::RngCore>(
...
+        name,
+        group_pub_key,
+        owner,
+        admin_vec,
+        member_vec,
+        needs_rotation,
+    )
+    .await?;
+
+    resp.try_into()
```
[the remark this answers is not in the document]
coltfred: Ya, I just think it's slightly more clear...

src/internal/rest.rs
coltfred: They stopped.

src/internal/rest.rs
coltfred: Yep.

src/internal/rest.rs
coltfred: Removed.

src/internal/rest.rs
coltfred: true. I was able to get rid of the indentation here as well.

src/internal/rest.rs
coltfred: done.

src/internal/rest.rs
coltfred: fixed.

src/internal/rest.rs
coltfred: fixed

Cargo.toml
coltfred: I think we do want that... Anything inside 1.x should be fine I would think?

src/internal/rest.rs
coltfred: Fixed.

src/internal/group_api/mod.rs
```diff
@@ -453,108 +450,95 @@ fn collect_admin_and_member_info<CR: rand::CryptoRng + rand::RngCore>(
...
-                group_pub_key,
-                owner,
-                admin_vec,
-                member_vec,
-                needs_rotation,
-            )
+) -> Result<GroupCreateResult, IronOxideErr> {
+    let user_ids_and_keys = user_api::user_key_list(auth, users_to_lookup).await?;
+    // this will occur when one of the UserIds cannot be found
+    let user_ids_and_keys = check_user_mismatch(users_to_lookup, user_ids_and_keys)?;
```
giarc3: I like this new layout much better. But what's the reason for this reassignment?

src/internal/rest.rs
```diff
@@ -8,21 +8,19 @@ use bytes::Bytes;
 use chrono::{DateTime, Utc};
 use percent_encoding::SIMPLE_ENCODE_SET;
 use reqwest::{
-    header::HeaderMap,
-    r#async::{Client as RClient, Request as ARequest},
+    header::{HeaderMap, HeaderValue, CONTENT_TYPE},
+    r#async::{Client as RClient, Request as ARequest, RequestBuilder},
```
clintfred: did you figure out that the async namespace is still the preferred way to import here? 
https://github.com/seanmonstar/reqwest/blob/80ba8cc15066c2a6d35d68b058e795be5e01cd94/src/lib.rs#L301 makes me think otherwise.

src/internal/rest.rs
```diff
@@ -186,6 +184,8 @@ pub struct SignatureUrlString(String);
 impl SignatureUrlString {
     pub fn new(encoded_full_url: &str) -> Result<SignatureUrlString, url::ParseError> {
         let parsed_url = Url::parse(encoded_full_url)?;
+        //The formatter for the query string is overfomatting compared to what we want, so we want to replace
```
clintfred: Do you think this is worth opening an issue on reqwest? Something like "Support configurable query string encoding". Perhaps if we reference our issue someone will point out a better way all together?

src/internal/rest.rs
coltfred: No, I just pushed the fix.

src/internal/group_api/mod.rs
coltfred: The check_user_mismatch takes a value and will either produce an error message from it or return it. It allows us to avoid cloning the entire map.

PR #89, opened 2019-12-05
Make the reqwest client static and reuse it for all requests

coltfred:
Currently I'm caching this globally, which allows us to share it for user create as well as for all Ironoxide instances. This has the following advantages:
- Shared even on user create, etc
- Shared even if the user creates new SDKs.


coltfred:
Note that I believe that `.clone` on the Reqwest Client just increments the Arc counter since the Client just holds an Arc. https://docs.rs/reqwest/0.10.0-alpha.2/src/reqwest/async_impl/client.rs.html#49

src/internal/mod.rs
```diff
@@ -24,7 +24,7 @@ use std::{
 
 pub mod document_api;
 pub mod group_api;
-mod rest;
+pub(crate) mod rest; //COLT: Feels a bit bad.
```
clintfred: an alternative would be to allow the internal API to generate the `IronCoreRequest` struct, allowing it to stay private to internal.

src/internal/mod.rs
```diff
@@ -38,7 +38,8 @@ lazy_static! {
         .to_string(),
         _ => "https://api.ironcorelabs.com/api/1/".to_string(),
     };
-    pub static ref OUR_REQUEST: IronCoreRequest = IronCoreRequest::new(URL_STRING.as_str());
+    // pub static ref OUR_REQUEST: IronCoreRequest =
```
clintfred: I wonder if it would be appropriate to open an issue on reqwest to better understand why this doesn't work?

coltfred:
At this point we don't have a great understanding about the contention issues in highly concurrent situations, so we're going to just close this. The reason we were trying it in the first place was because we wanted performance gains, but without that as a driving factor it doesn't seem like it's worth putting time into right now.

PR #90, opened 2019-12-05
Support group private key rotation

giarc3:
- Added request, response, and result types for `GroupUpdatePrivateKey`
- Added `current_key_id` to `GroupBasicApiResponse` and `GroupGetApiResponse`, but didn't include it in any result types
- Added `group_rotate_private_key()` to rotate the provided group's private key
- Moved `AugmentationFactor` from `internal::user_api::requests` to `internal::rest::json` so it can be accessed from `group_api` as well
- Added `group_rotation_needed()` to `PrivateKeyRotationCheckResult` to get back the groups that need rotation on `initialize_check_rotation()`
- Added tests for `group_rotate_private_key()` and `group_rotation_needed()`

TODO:
- [x] Changelog
- [x] Address reviewer TODO about a `rotate_all()` function
- [x] Add more tests that reviewers would like to include


src/internal/group_api/mod.rs
```diff
@@ -526,6 +530,82 @@ pub async fn group_create<CR: rand::CryptoRng + rand::RngCore>(
...
+        Some(user_id_strings) => user_id_strings
+            .into_iter()
+            .map(|string_id| string_id.to_owned().try_into())
+            .collect::<Result<Vec<_>, _>>(),
+        None => Err(IronOxideErr::NotGroupAdmin(group_id.to_owned())),
+    }?;
+    let (_, admins_with_keys) = get_user_keys(auth, &admins).await?;
+    let encrypted_group_key = group_info
+        .encrypted_private_key
+        .ok_or(IronOxideErr::NotGroupAdmin(group_id.to_owned()))?;
```
ernieturner: It seems like this check should be the first thing we do in this method. It's the quickest tell to know if the user is an admin of the group and therefore has the ability to do the rotation. As it is now, you're going to get the public keys for all of the admins before you know the the user can even rotate the key.

src/internal/group_api/mod.rs
```diff
@@ -526,6 +530,82 @@ pub async fn group_create<CR: rand::CryptoRng + rand::RngCore>(
...
+    let (old_plaintext, _) = transform::decrypt_plaintext(
+        &recrypt,
+        encrypted_group_key.try_into()?,
+        &device_private_key.recrypt_key(),
+    )?;
+    let old_group_private_key = recrypt.derive_private_key(&old_plaintext);
+    let (new_plaintext, new_group_private_key, _) = transform::gen_group_keys(recrypt)?;
+    let private_key_diff = old_group_private_key.augment_minus(&new_group_private_key);
+    let aug_factor = internal::AugmentationFactor(private_key_diff.into());
```
ernieturner: Shouldn't this be calling the `augment_private_key_with_retry` method that we do for user key rotation somewhere? I'm not seeing where this is checking that the new group private key nor the augmentation factor aren't all zeros.

src/internal/group_api/mod.rs
```diff
@@ -526,6 +530,82 @@ pub async fn group_create<CR: rand::CryptoRng + rand::RngCore>(
...
+) -> Result<GroupUpdatePrivateKeyResult, IronOxideErr> {
+    let group_info = group_get_request(&auth, group_id).await?;
+    let admins = match &group_info.admin_ids {
+        Some(user_id_strings) => user_id_strings
+            .into_iter()
+            .map(|string_id| string_id.to_owned().try_into())
+            .collect::<Result<Vec<_>, _>>(),
+        None => Err(IronOxideErr::NotGroupAdmin(group_id.to_owned())),
+    }?;
+    let (_, admins_with_keys) = get_user_keys(auth, &admins).await?;
```
ernieturner: I added a similar comment to Tonias PR, but we should probably check that we got all of the admins keys that we asked for and fail if we didn't

https://github.com/IronCoreLabs/ironweb/pull/65/files#r354552519

src/lib.rs
```diff
@@ -138,11 +139,15 @@ impl PrivateKeyRotationCheckResult {
...
+    pub fn group_rotation_needed(&self) -> Vec<GroupId> {
+        match &self.rotations_needed {
+            EitherOrBoth::Right(groups) | EitherOrBoth::Both(_, groups) => groups.to_vec(),
+            _ => vec![],
+        }
+    }
 
-    //    pub fn rotate_all(ironoxide: &IronOxide) -> (UserId, Vec<GroupId>) //TODO consider when group private key rotation is added
+    // TODO (reviewers): Is this something we want and I should implement as part of this ticket?
+    // pub fn rotate_all(ironoxide: &IronOxide) -> (UserId, Vec<GroupId>) //TODO consider when group private key rotation is added
```
ernieturner: Unless this is something that NCLH requires, I would say we punt on implementing this.

src/internal/group_api/mod.rs
```diff
@@ -526,6 +530,82 @@ pub async fn group_create<CR: rand::CryptoRng + rand::RngCore>(
...
+    let encrypted_group_key = group_info
+        .encrypted_private_key
+        .ok_or(IronOxideErr::NotGroupAdmin(group_id.to_owned()))?;
+    let (old_plaintext, _) = transform::decrypt_plaintext(
```
coltfred: We should write a version of this that returns (Plaintext, PrivateKey) instead of the symmetric key. Today it's confusing that we throw the symmetric key away and then hash to get the private key. This will simplify `group_add_admins` as well.

src/internal/group_api/mod.rs
```diff
@@ -526,6 +530,82 @@ pub async fn group_create<CR: rand::CryptoRng + rand::RngCore>(
...
+    let old_group_private_key = recrypt.derive_private_key(&old_plaintext);
+    let (new_plaintext, new_group_private_key, _) = transform::gen_group_keys(recrypt)?;
+    let private_key_diff = old_group_private_key.augment_minus(&new_group_private_key);
+    let aug_factor = internal::AugmentationFactor(private_key_diff.into());
+    let (errors, admin_info) = transform::encrypt_to_with_key(
+        recrypt,
+        &new_plaintext,
+        &auth.signing_private_key().into(),
+        admins_with_keys,
+    );
```
coltfred: It seems like 564-578 would make a good function that would be unit testable without the server.

src/internal/group_api/mod.rs
```diff
@@ -526,6 +530,82 @@ pub async fn group_create<CR: rand::CryptoRng + rand::RngCore>(
...
+    // This will return an Err for the function if encrypting to any of the admins failed.
+    errors
+        .into_iter()
+        .map(|(_, error)| Err(error.into()))
+        .collect::<Result<(), IronOxideErr>>()?;
+    let new_group_admins = admin_info
```
coltfred: This code is duplicated from group_add_admin. Could make a function that takes `Vec<(WithKey<A>, EncryptedValue)>` and returns `Result<(A,PublicKey,IronoxideEncryptedValue)>`

src/internal/group_api/requests.rs
```diff
@@ -11,13 +12,12 @@ use crate::internal::{
     IronOxideErr, RequestAuth, RequestErrorCode, SchnorrSignature,
 };
 use chrono::{DateTime, Utc};
+use internal::auth_v2::AuthV2Builder;
```
coltfred: Seems like this might be able to merged with the above import? (I know it wasn't your fault).

src/lib.rs
```diff
@@ -138,11 +139,15 @@ impl PrivateKeyRotationCheckResult {
...
-    //    }
+    pub fn group_rotation_needed(&self) -> Vec<GroupId> {
+        match &self.rotations_needed {
+            EitherOrBoth::Right(groups) | EitherOrBoth::Both(_, groups) => groups.to_vec(),
+            _ => vec![],
+        }
+    }
 
-    //    pub fn rotate_all(ironoxide: &IronOxide) -> (UserId, Vec<GroupId>) //TODO consider when group private key rotation is added
+    // TODO (reviewers): Is this something we want and I should implement as part of this ticket?
```
coltfred: Seems like a convenience we should write later. I'd write a ticket and leave it.

src/internal/group_api/mod.rs
```diff
@@ -526,6 +530,82 @@ pub async fn group_create<CR: rand::CryptoRng + rand::RngCore>(
...
+
+/// Rotate the group's private key. The public key for the group remains unchanged.
+pub async fn group_rotate_private_key<CR: rand::CryptoRng + rand::RngCore>(
+    recrypt: &Recrypt<Sha256, Ed25519, RandomBytes<CR>>,
+    group_id: &GroupId,
+    auth: &RequestAuth,
+    device_private_key: &PrivateKey,
+) -> Result<GroupUpdatePrivateKeyResult, IronOxideErr> {
+    let group_info = group_get_request(&auth, group_id).await?;
+    let admins = match group_info.admin_ids {
```
coltfred: Would you be willing to explore making the vecs in the request structure just have UserIds in them?

src/lib.rs
```diff
@@ -164,25 +169,45 @@ pub fn initialize_check_rotation(device_context: &DeviceContext) -> Result<InitA
...
+    rotations_needed: EitherOrBoth<UserId, Vec1<GroupId>>,
+    ironoxide: IronOxide,
+) -> InitAndRotationCheck {
+    InitAndRotationCheck::RotationNeeded(
+        ironoxide,
+        PrivateKeyRotationCheckResult { rotations_needed },
+    )
+}
+
 async fn initialize_check_rotation_async(
```
clintfred: Now that this is an `async fn`, could probably remove the `_async` suffix

src/lib.rs
```diff
@@ -164,25 +169,45 @@ pub fn initialize_check_rotation(device_context: &DeviceContext) -> Result<InitA
...
+    let maybe_groups_needing_rotation = Vec1::try_from_vec(groups_needing_rotation).ok();
+
+    Ok(
+        match (curr_user.needs_rotation(), maybe_groups_needing_rotation) {
+            (false, None) => InitAndRotationCheck::NoRotationNeeded(ironoxide),
+            (true, None) => form_rotation_needed(Left(account_id), ironoxide),
+            (false, Some(groups)) => form_rotation_needed(Right(groups), ironoxide),
+            (true, Some(groups)) => form_rotation_needed(Both(account_id, groups), ironoxide),
+        },
+    )
```
clintfred: I think this is a good candidate for a private function that can be unit tested.

src/internal/group_api/mod.rs
```diff
@@ -526,6 +530,82 @@ pub async fn group_create<CR: rand::CryptoRng + rand::RngCore>(
...
+}
+
+impl GroupUpdatePrivateKeyResult {
+    /// True if this group's private key requires additional rotation
+    pub fn needs_rotation(&self) -> bool {
+        self.needs_rotation
+    }
+}
+
+/// Rotate the group's private key. The public key for the group remains unchanged.
```
clintfred: Please write some more docs on this `pub fn`. I realize that user public key rotation doesn't have good docs (on me), but since this is our internal API, we should document these.

src/lib.rs
clintfred: I agree that delaying implementation here is fine. I think the most common use case for initializing with rotation is that you will want to just do any rotations that are needed, so it would simplify the caller's code significantly if we provide this.

src/lib.rs
```diff
@@ -138,11 +139,15 @@ impl PrivateKeyRotationCheckResult {
         }
     }
 
-    //    pub fn group_rotation_needed() -> Option<Vec<GroupId>> {
-    //        unimplemented!()
-    //    }
+    pub fn group_rotation_needed(&self) -> Vec<GroupId> {
```
clintfred: I think the return type here should match `rotations_needed`. That is field is `pub` so that callers can directly match on it if they want to. 

I think returning `Option<Vec1<GroupId>>` here would be good.

src/lib.rs
giarc3: How would you break this up so that the private function doesn't need an ironoxide, so that I can write a unit test on it?

src/internal/group_api/mod.rs
ernieturner: This is roughly what Tonia is doing in her PR as well. We only return the encrypted group private key if the user is an admin, so it's the best way we have right now to check that.

src/internal/group_api/mod.rs
ernieturner: I agree the chance is pretty small, I just figured that if we were doing the retry logic for user rotation we should do the same for group rotation here as well.

src/internal/group_api/mod.rs
ernieturner: Colt commented on Tonias and corrected me that the server will reject if the list of keys isn't the same, so I was wrong here. It doesn't hurt anything to check the list, but it's not necessary. So up to you to either leave it in or remove this. Sorry I led you astray.

src/internal/group_api/mod.rs
giarc3: I added a unit test that verifies that the augmentation works for all the admins.

src/lib.rs
giarc3: I needed to make some things pub(crate) so that I could build them from scratch for the test, but I was able to add a unit test for this function.

src/internal/user_api/mod.rs
```diff
@@ -121,10 +121,10 @@ pub struct DeviceAdd {
 /// IDs and public key for existing user on verify result
 #[derive(Debug)]
 pub struct UserResult {
-    account_id: UserId,
-    segment_id: usize,
-    user_public_key: PublicKey,
-    needs_rotation: bool,
+    pub(crate) account_id: UserId,
```
clintfred: See comment on GroupMetaResult

src/lib.rs
```diff
@@ -165,25 +176,50 @@ pub fn initialize_check_rotation(device_context: &DeviceContext) -> Result<InitA
         .block_on(initialize_check_rotation_async(device_context))
 }
 
+// Finds the groups that the caller is an admin of that need rotation and
```
clintfred: `///`

src/lib.rs
clintfred: Good point. We have some work to do to separate sync and async APIs. We can fix this up at that point.

src/crypto/transform.rs
```diff
@@ -18,26 +18,36 @@ pub fn generate_new_doc_key<CR: rand::CryptoRng + rand::RngCore>(
...
     let plaintext = recrypt.gen_plaintext();
     let priv_key = recrypt.derive_private_key(&plaintext);
     let pub_key = recrypt.compute_public_key(&priv_key)?;
-
     Ok((plaintext, priv_key.into(), pub_key.into()))
 }
 
-/// Decrypt the provided encrypted plaintext and return both the plaintext and the symmetric key that
+/// Decrypt the provided encrypted plaintext and return the symmetric key that is derived from it.
+pub fn decrypt_symmetric_key<CR: rand::CryptoRng + rand::RngCore>(
```
clintfred: suggestion: `decrypt_as_symmetric_key`

src/internal/mod.rs
```diff
@@ -1027,6 +1053,43 @@ pub(crate) mod test {
             unimplemented!()
         }
     }
+    // impl CryptoOps for MockKeyGenOps {
```
clintfred: What's going on with this test?

src/internal/mod.rs
```diff
@@ -729,6 +736,25 @@ fn augment_private_key_with_retry<R: KeyGenOps>(
     aug_private_key().or_else(|_| aug_private_key())
 }
 
+// Subtracts a generated private key from the provided PrivateKey, returning
+// the result and the plaintext associated with the generated key.
+// There is a very small chance that the generated private key could not be compatible with
+// the given PrivateKey, so we retry once internally before giving the caller an error.
+fn gen_plaintext_and_aug_with_retry<R: recrypt::api::CryptoOps + KeyGenOps>(
```
clintfred: Is there an opportunity to combine this with the other augmentation function? Maybe tying augmentation to PrivateKey was the wrong thing? Wondering if you have thoughts here. If we're going to address, we probably should do so in another PR.

src/internal/mod.rs
giarc3: It's currently commented out as I don't know how to mock the CryptoOps needed for `transform::gen_group_keys`. I will either need help writing it, or suggestions on an alternative.

src/internal/group_api/mod.rs
```diff
@@ -948,4 +1048,57 @@ mod test {
         assert_eq!(admin_info.len(), 1);
         Ok(())
     }
+
+    #[test]
+    fn augment_admin_list() -> Result<(), IronOxideErr> {
```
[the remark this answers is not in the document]
giarc3: Added

src/internal/group_api/mod.rs
```diff
@@ -120,14 +120,14 @@ impl GroupListResult {
...
 pub struct GroupMetaResult {
-    id: GroupId,
-    name: Option<GroupName>,
-    group_master_public_key: PublicKey,
-    is_admin: bool,
-    is_member: bool,
-    created: DateTime<Utc>,
-    updated: DateTime<Utc>,
-    needs_rotation: Option<bool>,
+    pub(crate) id: GroupId,
```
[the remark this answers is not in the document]
giarc3: Put creator functions in the `test` module to avoid making these `pub(crate)`.

src/internal/group_api/mod.rs
```diff
@@ -594,13 +559,79 @@ pub async fn group_rotate_private_key<CR: rand::CryptoRng + rand::RngCore>(
                     encrypted_msg,
                 })
         })
-        .collect::<Result<Vec<GroupAdmin>, IronOxideErr>>()?;
+        .collect::<Result<Vec<GroupAdmin>, IronOxideErr>>()
+}
+
+// Decrypts the group's private key, generates a new private key and plaintext,
+// computes the difference between the old and new private keys, and encrypts the new plaintext to each group admin.
```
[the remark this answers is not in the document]
giarc3: Fixed

src/internal/group_api/mod.rs
```diff
@@ -594,13 +559,79 @@ pub async fn group_rotate_private_key<CR: rand::CryptoRng + rand::RngCore>(
...
+fn generate_aug_and_admins<CR: rand::CryptoRng + rand::RngCore>(
+    recrypt: &Recrypt<Sha256, Ed25519, RandomBytes<CR>>,
+    auth: &RequestAuth,
```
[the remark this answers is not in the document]
giarc3: Good point. That simplifies the unit test for it as well. Done.

src/internal/user_api/mod.rs
giarc3: Done

src/lib.rs
giarc3: Fixed

src/crypto/transform.rs
giarc3: Changed. I also changed the companion function to `decrypt_as_private_key` to match.

src/internal/mod.rs
giarc3: It's now implemented. I left a comment on why I was unable to test that it retries if the private key is zero, but that should be covered in the other unit test that calls the same `augment` function.

src/internal/mod.rs
giarc3: They likely could in the future, but since one requires `KeyGenOps` to make an `AugmentationFactor` and the other requires `CryptoOps` to generate a plaintext/derive its private key, I'd prefer to leave them separate in this PR.

src/internal/group_api/mod.rs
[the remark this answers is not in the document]
giarc3: Put creator functions in the test module to avoid making these pub(crate).

src/internal/group_api/mod.rs
```diff
@@ -526,6 +530,112 @@ pub async fn group_create<CR: rand::CryptoRng + rand::RngCore>(
...
+
+impl GroupUpdatePrivateKeyResult {
+    /// True if this group's private key requires additional rotation
+    pub fn needs_rotation(&self) -> bool {
+        self.needs_rotation
+    }
+}
+
+/// Maps the successful results of `transform::encrypt_to_with_key()` into a vector of GroupAdmins
+fn collect_request_admins(
```
ernieturner: Not a huge deal, but the name of this function made me confused for quite a while. Assuming that I understand what it does, maybe `collect_group_admin_keys` or similar? I think I'm mostly just confused about what "request_admins" means.

giarc3: Sure, I can make that change. "request_admins" was just referring to the requests::GroupAdmin that we needed a vector of when sending the request.

Issue #84, opened 2019-12-03
Reuse Tokio Runtime for DeviceContext auth'd operations


clintfred:
As an ironoxide user
I want SDK functions to execute quickly without significant uncessary overhead

As part of some [experimentation with flame](https://github.com/IronCoreLabs/ironoxide/tree/flame-instrument) I pulled the Tokio Runtime out of each public API. I don't really know if that had a noticeable effect on performance, but it certainly felt cleaner. I had previously suspected that creating/destroying a Runtime might have significant overhead. 

We have decided not to leave flame annotations in our normal code due to limitations with concurrent operations. The tokio runtime work should be pulled out of that branch and PR'd separately. 

AC:
* [ ] All DeviceContext authenticated SDK functions use a shared `Runtime` struct
* [ ] cached Runtime allows for true multithreaded behavior

Non AC: 
* JWT auth'd SDK functions are using a shared `Runtime`. We may choose to try to address this at some point, but might want to use a more sophisticated pattern (like what `reqwest` does).

clintfred:
Increased estimate to 3 due to increased scope of trying to provide better default config to tokio

PR #91, opened 2019-12-05
[#84] Reuse tokio runtime for device auth'd SDK calls

clintfred:
see #84 

## TODO
* [x] changelog entry

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -194,8 +194,9 @@ impl IronOxide {
 
     /// Create an IronOxide instance. Depends on the system having enough entropy to seed a RNG.
     fn create(curr_user: &UserResult, device_context: &DeviceContext) -> IronOxide {
-        let rt = tokio::runtime::Builder::new()
-            //            .core_threads(8)  // default is num_cpus
+        // create a tokio runtime with the default number of core threads (num of cores on a machine)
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -193,6 +194,12 @@ impl IronOxide {
 
     /// Create an IronOxide instance. Depends on the system having enough entropy to seed a RNG.
     fn create(curr_user: &UserResult, device_context: &DeviceContext) -> IronOxide {
+        let rt = tokio::runtime::Builder::new()
+//            .core_threads(8)  // default is num_cpus
```
coltfred: I think a comment on the `let rt` with something to the effect of `This runtime should have the core_threads set to the number of cpus on the machine, but with more blocking threads since we expect the SDK methods to block frequently on network`

Then you can remove this commented out line.

src/lib.rs
```diff
@@ -202,6 +209,7 @@ impl IronOxide {
                 BYTES_BEFORE_RESEEDING,
                 EntropyRng::new(),
             )),
+            runtime: rt,
```
coltfred: Nitpick: Is there a reason not to name the above variable 'runtime' to make this shorter?

src/lib.rs
clintfred: done

coltfred:
Don't forget the changelog entry.

PR #94, opened 2019-12-12
Implement rotate_all() for PrivateKeyRotationCheckResult

giarc3:
These changes are so that ironoxide-java can make its `initialize_and_rotate()` function include all groups that need rotation. 
I also noticed that the GroupUpdatePrivateKeyResult only returning `needs_rotation` makes sense when you're calling the function yourself one-by-one, but a `rotate_all()` will return a vector of results, so it should include the `GroupId` to relate the result to the group.

### DIFF
--- a/src/internal/group_api/requests.rs
+++ b/src/internal/group_api/requests.rs
@@ -290,11 +290,12 @@ pub mod group_update_private_key {
         needs_rotation: bool,
     }
 
-    impl From<GroupUpdatePrivateKeyResponse> for GroupUpdatePrivateKeyResult {
-        fn from(resp: GroupUpdatePrivateKeyResponse) -> Self {
+    impl From<(GroupUpdatePrivateKeyResponse, GroupId)> for GroupUpdatePrivateKeyResult {
+        fn from(resp_and_id: (GroupUpdatePrivateKeyResponse, GroupId)) -> Self {
             // don't expose the current_key_id to the outside world until we need to
             GroupUpdatePrivateKeyResult {
-                needs_rotation: resp.needs_rotation,
+                id: resp_and_id.1,
+                needs_rotation: resp_and_id.0.needs_rotation,
             }
         }
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -156,6 +156,10 @@ impl PrivateKeyRotationCheckResult {
         }
     }
 
+    /// Rotate the private key of the calling user and all groups they are an administrator of where needs_rotation is true.
+    /// Note that this function has the potential to take much longer than other functions, as rotation will be done
+    /// individually on each user/group. If rotation is only needed for a specific group, it is strongly recommended
+    /// to call group_rotate_private_key() instead.
     pub fn rotate_all(
         &self,
         ironoxide: &IronOxide,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -159,7 +159,11 @@ impl PrivateKeyRotationCheckResult {
     /// Rotate the private key of the calling user and all groups they are an administrator of where needs_rotation is true.
     /// Note that this function has the potential to take much longer than other functions, as rotation will be done
     /// individually on each user/group. If rotation is only needed for a specific group, it is strongly recommended
-    /// to call group_rotate_private_key() instead.
+    /// to call [user_rotate_private_key()](user\/trait.UserOps.html#tymethod.user_rotate_private_key) or
+    /// [group_rotate_private_key()](group\/trait.GroupOps.html#tymethod.group_rotate_private_key) instead.
+    /// # Arguments
+    /// - `ironoxide` - IronOxide used to make authenticated requests for the calling user
+    /// - `password` - Password to unlock the current user's user master key
     pub fn rotate_all(
         &self,
         ironoxide: &IronOxide,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -73,8 +73,8 @@ pub mod policy;
 pub mod prelude;
 
 use crate::internal::{
-    group_api::GroupId,
-    user_api::{UserId, UserResult},
+    group_api::{GroupId, GroupUpdatePrivateKeyResult},
+    user_api::{UserId, UserResult, UserUpdatePrivateKeyResult},
 };
 pub use crate::internal::{
     DeviceContext, DeviceSigningKeyPair, IronOxideErr, KeyPair, PrivateKey, PublicKey,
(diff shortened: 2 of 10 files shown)


src/lib.rs
```diff
@@ -142,21 +142,65 @@ pub struct PrivateKeyRotationCheckResult {
...
+    pub fn group_rotation_needed(&self) -> Option<&Vec1<GroupId>> {
         match &self.rotations_needed {
-            EitherOrBoth::Right(groups) | EitherOrBoth::Both(_, groups) => Some(groups.clone()),
+            EitherOrBoth::Right(groups) | EitherOrBoth::Both(_, groups) => Some(groups),
             _ => None,
         }
     }
 
-    // pub fn rotate_all(ironoxide: &IronOxide) -> (UserId, Vec<GroupId>) //TODO consider when group private key rotation is added
+    pub fn rotate_all(
```
ernieturner: How does this method end up getting documented? We should have some fairly big warnings about how long this method might take. It's probably unlikely, but if there was a half-dozen groups to rotate and each one had a dozen admins in it, this method will take a looooong time.

giarc3: I added a comment that says that this could take a long time, and strongly recommending using the regular function for individual groups if possible. Let me know if you would change/add to it.

src/lib.rs
```diff
@@ -142,21 +142,69 @@ pub struct PrivateKeyRotationCheckResult {
...
+    )> {
+        use futures::future::OptionFuture;
+        let user_future = self.user_rotation_needed().and(Some(
+            crate::internal::user_api::user_rotate_private_key(
+                &ironoxide.recrypt,
+                password.try_into()?,
+                ironoxide.device().auth(),
+            ),
+        ));
+        let group_futures = match self.group_rotation_needed() {
```
coltfred: You should be able to map to this instead.

tests/group_ops.rs
```diff
@@ -154,6 +154,66 @@ fn group_rotate_private_key_non_admin() -> Result<(), IronOxideErr> {
...
+    let jwt = gen_jwt(Some(account_id.id())).0;
+    IronOxide::user_create(&jwt, USER_PASSWORD, &UserCreateOpts::new(true))?;
+    let device = IronOxide::generate_new_device(
+        &gen_jwt(Some(account_id.id())).0,
+        USER_PASSWORD,
+        &Default::default(),
+    )?;
+    let creator_sdk = ironoxide::initialize(&device)?;
+    // making non-default groups so I can specify needs_rotation of true
+    let group_create1 = creator_sdk.group_create(&GroupCreateOpts::new(
```
coltfred: `let group_create1 = GroupCreateOpts { needs_rotation: true, ..Default::default() };`

tests/group_ops.rs
```diff
@@ -154,6 +154,66 @@ fn group_rotate_private_key_non_admin() -> Result<(), IronOxideErr> {
...
+        None,
+        true,
+        true,
+        None,
+        vec![],
+        vec![],
+        true,
+    ))?;
+    assert_eq!(group_create1.needs_rotation(), Some(true));
+    let group_create2 = creator_sdk.group_create(&GroupCreateOpts::new(
```
coltfred: Ditto here.

tests/group_ops.rs
giarc3: Sadly this won't work because the fields of the struct are private

tests/group_ops.rs
giarc3: Ditto here.

src/lib.rs
giarc3: I changed this and the user_future both to map to be more readable

src/lib.rs
```diff
@@ -142,21 +142,67 @@ pub struct PrivateKeyRotationCheckResult {
...
+                .into_iter()
+                .map(|group_id| {
+                    crate::internal::group_api::group_rotate_private_key(
+                        &ironoxide.recrypt,
+                        ironoxide.device().auth(),
+                        &group_id,
+                        ironoxide.device().device_private_key(),
+                    )
+                })
+                .collect::<Vec<_>>();
```
clintfred: I've noticed you having to use the turbofish quite a bit for this purpose. Note that itertools has a `collect_vec()` for this purpose. Just an option.

src/lib.rs
clintfred: Just a thought: If the app/runtime has the ability to execute the rotations in parallel, it probably won't be that bad.

src/lib.rs
```diff
@@ -142,21 +142,67 @@ pub struct PrivateKeyRotationCheckResult {
...
+            EitherOrBoth::Right(groups) | EitherOrBoth::Both(_, groups) => Some(groups),
             _ => None,
         }
     }
 
-    // pub fn rotate_all(ironoxide: &IronOxide) -> (UserId, Vec<GroupId>) //TODO consider when group private key rotation is added
+    /// Rotate the private key of the calling user and all groups they are an administrator of where needs_rotation is true.
+    /// Note that this function has the potential to take much longer than other functions, as rotation will be done
+    /// individually on each user/group. If rotation is only needed for a specific group, it is strongly recommended
+    /// to call group_rotate_private_key() instead.
```
clintfred: Need to document signature of this function.

I'm not sure pointing the caller just to `group_rotate_private_key` is a good idea as this rotates both user and group private keys.

src/internal/group_api/requests.rs
```diff
@@ -290,11 +290,12 @@ pub mod group_update_private_key {
         needs_rotation: bool,
     }
 
-    impl From<GroupUpdatePrivateKeyResponse> for GroupUpdatePrivateKeyResult {
-        fn from(resp: GroupUpdatePrivateKeyResponse) -> Self {
+    impl From<(GroupUpdatePrivateKeyResponse, GroupId)> for GroupUpdatePrivateKeyResult {
```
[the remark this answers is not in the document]
giarc3: I agree that that would make sense/be cleaner, but I would say that it should be a separate ticket for ironcore-id and this is a reasonable solution for now. Do you disagree?

src/lib.rs
giarc3: I think I prefer using the turbofish since the internal type inference works well and it is consistent with the collects into non-vec types

src/lib.rs
giarc3: I documented the parameters, called out `user_rotate_private_key` as well, and added links to the two functions.

src/lib.rs
clintfred: That's fine. Just providing another option.

Issue #93, opened 2019-12-11
Create async public API


clintfred:
As an ironoxide user
I want an `async fn` based API so that I can easily asynchronously compose ironoxide calls with other operations.

AC:
* [ ] The current sync API will be deprecated (for now) and an async one will replace it
* [ ] All references to Runtime (and tokio) will be removed from ironoxide

Currently our sync API cannot be composed with external async code as we hold an internal tokio runtime/reactor, and there is a conflict of some sort.

```
'nested block_on: EnterError { reason: "attempted to run an executor while another executor is already running" }', src/libcore/result.rs:1165:5
stack backtrace:
<snip>
16: tokio::runtime::threadpool::Runtime::block_on
```

clintfred:
This does not build on 1.40 stable, but it does build on beta...

PR #97, opened 2019-12-19
[#93] Make async api the default

clintfred:
See #93 

We've decided that for now we are going to only provide an async API as this simplifies things given our internal API is async. If we need a sync API again in the future, it could be added in a separate name space, perhaps with a feature flag.

Obviously this is a huge breaking change, so we're merging this into a feature branch for now.

## TODO
* [x] changelog entry
* [ ] ~make sure this works on stable~ (do this prior to merging to master)

### DIFF
--- a/src/document/mod.rs
+++ b/src/document/mod.rs
@@ -88,12 +88,13 @@ impl Default for DocumentEncryptOpts {
     }
 }
 
+#[async_trait]
 pub trait DocumentOps {
     /// List all of the documents that the current user is able to decrypt.
     ///
     /// # Returns
     /// `DocumentListResult` struct with vec of metadata about each document the user can decrypt.
-    fn document_list(&self) -> Result<DocumentListResult>;
+    async fn document_list(&self) -> Result<DocumentListResult>;
 
     /// Get the metadata for a specific document given its ID.
     ///
@@ -102,7 +103,7 @@ pub trait DocumentOps {
     ///
     /// # Returns
     /// `DocumentMetadataResult` with details about the requested document.
-    fn document_get_metadata(&self, id: &DocumentId) -> Result<DocumentMetadataResult>;
+    async fn document_get_metadata(&self, id: &DocumentId) -> Result<DocumentMetadataResult>;
 
     /// Attempt to parse the document ID out of an encrypted document.
     ///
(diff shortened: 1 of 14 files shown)


tests/document_ops.rs
```diff
@@ -504,267 +518,291 @@ fn doc_create_must_grant() -> Result<(), IronOxideErr> {
...
-    let mut joined_count = 0;
-    for t in threads {
-        t.join().expect("couldn't join");
-        joined_count += 1;
-    }
-
-    assert_eq!(joined_count, 10);
-    Ok(())
-}
+//#[tokio::test]
```
clintfred: Since we have throughput test/benches coming, I think I'm going to remove this test completely. Comment if you disagree.

src/document/mod.rs
```diff
@@ -235,85 +237,81 @@ impl DocumentOps for crate::IronOxide {
...
+        document_api::encrypt_document(
             self.device.auth(),
             &self.recrypt,
             &self.user_master_pub_key,
             &self.rng,
             document_data,
-            encrypt_opts.id,
-            encrypt_opts.name,
+            maybe_doc_id,
+            maybe_doc_name,
```
giarc3: What's the reasoning behind this change? To me it looks the same but with an extra `clone()` thrown onto the document id and name

tests/group_ops.rs
```diff
@@ -154,81 +169,86 @@ fn group_rotate_private_key_non_admin() -> Result<(), IronOxideErr> {
...
     Ok(())
 }
 
-#[test]
-fn group_get_metadata() -> Result<(), IronOxideErr> {
-    let admin_sdk = initialize_sdk()?;
-    let member_sdk = initialize_sdk()?;
-    let nonmember_sdk = initialize_sdk()?;
+#[tokio::test]
+async fn group_get_metadata() -> Result<(), IronOxideErr> {
```
giarc3: I assume that this function would be a good place to use `join`s instead of back-to-back `await`s. This might not be the PR to do so, but this function in particular looks like we could speed it up.

src/document/mod.rs
clintfred: I was getting compiler errors about partial moves. This was my way around that.

tests/group_ops.rs
clintfred: Fair observation. I'm going for a 1-1 conversion in this PR, however and these were being done in serial before.

tests/document_ops.rs
coltfred: Feel free.

PR #98, opened 2019-12-19
Remove device_id from DeviceContext

giarc3:
- Removed device_id from `RequestAuth`
- Removed device_id from `DeviceContext::new()`

TODO:
- [x] Changelog

### DIFF
--- a/src/internal/mod.rs
+++ b/src/internal/mod.rs
@@ -5,7 +5,7 @@
 use crate::internal::{
     group_api::GroupId,
     rest::{Authorization, IronCoreRequest, SignatureUrlString},
-    user_api::{DeviceId, UserId},
+    user_api::UserId,
 };
 use chrono::{DateTime, Utc};
 use log::error;
@@ -251,7 +251,6 @@ pub mod auth_v2 {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct RequestAuth {
-    device_id: DeviceId,
     ///The users given id, which uniquely identifies them inside the segment.
     account_id: UserId,
     ///The segment_id for the above user.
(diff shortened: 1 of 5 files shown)


src/internal/mod.rs
```diff
@@ -828,21 +821,19 @@ pub(crate) mod test {
...
         let context = DeviceContext::new(
-            314.try_into()?,
             "account_id".try_into()?,
             22,
             priv_key.into(),
             DeviceSigningKeyPair::from(dev_keys),
         );
         let json = serde_json::to_string(&context).unwrap();
-        let expect_json = r#"{"deviceId":314,"accountId":"account_id","segmentId":22,"signingPrivateKey":"AQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQGKiOPddAnxlf1S2y08ul1yymcJvx2UEhvzdIgBtA9vXA==","devicePrivateKey":"bzb0Rlg0u7gx9wDuk1ppRI77OH/0ferXleenJ3Ag6Jg="}"#;
+        let expect_json = r#"{"accountId":"account_id","segmentId":22,"signingPrivateKey":"AQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQGKiOPddAnxlf1S2y08ul1yymcJvx2UEhvzdIgBtA9vXA==","devicePrivateKey":"bzb0Rlg0u7gx9wDuk1ppRI77OH/0ferXleenJ3Ag6Jg="}"#;
```
ernieturner: Just curious, would this fail if the deviceId is in this json, or does serde just ignore it?

giarc3: This string is just used to check the DeviceContext serialization, so it needs to be removed here since it's no longer in the struct. But if I deserialized this string into a new DeviceContext, it would ignore the deviceId and work fine.

PR #101, opened 2020-01-06
Update reqwest to 0.10.0

giarc3:
- Update `bytes` to 0.5
- Update `reqwest` to 0.10
- Update `tokio` to 0.2.6
- Update `protobuf` to 2.10
- Standardized dependencies to use `~` to try to avoid our errors of misunderstanding `~` vs `^`

### DIFF
--- a/src/internal/document_api/mod.rs
+++ b/src/internal/document_api/mod.rs
@@ -768,12 +768,12 @@ impl TryFrom<&EncryptedDek> for EncryptedDekP {
             } => {
                 let mut proto_edek_data = EncryptedDekDataP::default();
 
-                proto_edek_data.set_encryptedBytes(encrypted_message.bytes()[..].into());
+                proto_edek_data.set_encryptedBytes(encrypted_message.bytes().to_vec().into());
                 proto_edek_data
                     .set_ephemeralPublicKey(PublicKey::from(ephemeral_public_key).into());
-                proto_edek_data.set_signature(signature.bytes()[..].into());
-                proto_edek_data.set_authHash(auth_hash.bytes()[..].into());
-                proto_edek_data.set_publicSigningKey(public_signing_key.bytes()[..].into());
+                proto_edek_data.set_signature(signature.bytes().to_vec().into());
+                proto_edek_data.set_authHash(auth_hash.bytes().to_vec().into());
+                proto_edek_data.set_publicSigningKey(public_signing_key.bytes().to_vec().into());
 
                 Ok(proto_edek_data)
             }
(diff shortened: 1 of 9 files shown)


src/lib.rs
```diff
@@ -283,8 +283,9 @@ impl IronOxide {
         // create a tokio runtime with the default number of core threads (num of cores on a machine)
         // and an elevated number of blocking_threads as we expect heavy concurrency to be network-bound
         let runtime = tokio::runtime::Builder::new()
-            .blocking_threads(250) // most all SDK methods will block on the network
-            .keep_alive(None)
+            .threaded_scheduler()
+            .enable_all()
+            .max_threads(250) // most all SDK methods will block on the network
```
ernieturner: Maybe just for me, but it might be nice to comment each of these lines to explain what they mean. Or link to docs about it? Feel free to ignore if this is something obvious.

coltfred: Commenting the lines for now seems reasonable to me. @giarc3, would you please?

tests/common/mod.rs
```diff
@@ -45,27 +45,24 @@ lazy_static! {
...
     };
     static ref CONFIG: Config = {
-        use std::{error::Error, fs::File, io::Read};
+        use std::{fs::File, io::Read};
         let mut file = File::open(IRONCORE_CONFIG_PATH.1.clone()).unwrap_or_else(|err| {
             panic!(
                 "Failed to open config file ({}) with error '{}'",
-                IRONCORE_CONFIG_PATH.0,
-                err.description()
+                IRONCORE_CONFIG_PATH.0, err
```
coltfred: Worth noting that these changes were due to description being soft deprecated, not the version upgrade.

src/lib.rs
giarc3: Done

src/internal/document_api/mod.rs
```diff
@@ -768,12 +768,12 @@ impl TryFrom<&EncryptedDek> for EncryptedDekP {
             } => {
                 let mut proto_edek_data = EncryptedDekDataP::default();
 
-                proto_edek_data.set_encryptedBytes(encrypted_message.bytes()[..].into());
+                proto_edek_data.set_encryptedBytes(encrypted_message.bytes().to_vec().into());
```
clintfred: Was this due to a change in the `bytes` dependency? Can you explain what's going on here in some detail?

giarc3: They were already doing a `to_vec` internally, but they've now removed the From implementation for byte slices. So I just had to do the `to_vec` myself since they have a From for that.

clintfred: Understood. Thanks.

PR #104, opened 2020-01-09
Split functionality into Sync and Async

giarc3:
- Created the feature flag `"blocking"` that enables blocking versions of the functions.
- Moved the `tokio` dependency to be optional based on the flags
- Created a `blocking.rs` that contains a new struct `BlockingIronOxide`. This struct contains a regular IronOxide and a tokio Runtime.
- Implemented SDK functions for `BlockingIronOxide`. All of these simply use the Runtime to block on the asynchronous functions. This will make them very maintainable.
- Re-implemented `initialize()` and `initialize_check_rotation()`, again calling their async counterparts.
- Moved `rotate_all()` from `PrivateKeyRotationCheckResult` to `IronOxide`. It was always strange to have to pass an IronOxide to the function instead of just the result from `InitAndRotationCheck::RotationNeeded`, and moving it made it possible to share it between the sync/async versions.
- Added `test/blocking_ops.rs` to do minimal testing of the blocking functions. The tests call at least one function from every type of `*Ops` as well as `initialize()` and `initialize_check_rotation()`. These tests are run with `cargo t --features "blocking"`. Unfortunately, running regular `cargo t` will include `running 0 tests` from the new file, as I can't figure out how to make the whole file conditional under the feature flag.
- Minor cleanup/re-arranging of related files.


src/blocking.rs
```diff
@@ -0,0 +1,646 @@
+//! A blocking version of the SDK.
```
clintfred: I have mixed feelings about all blocking functionality living in this file. Did you consider breaking it up into separate, smaller files?

src/blocking.rs
```diff
@@ -0,0 +1,646 @@
...
+            BlockingIronOxide {
+                ironoxide: io,
+                runtime: rt,
+            },
+            rot,
+        ),
+    })
+}
+
+pub trait DocumentOps {
```
clintfred: Duplicating these traits and all the docs to remove the `async` from the functions seems sub-optimal. Assuming there aren't better options, we could use a procedural macro to generate these traits. Maybe someone else can think of a better option?

src/lib.rs
```diff
@@ -105,17 +109,17 @@ pub struct IronOxide {
 }
 
 /// Result of calling `initialize_check_rotation`
-pub enum InitAndRotationCheck {
+pub enum InitAndRotationCheck<T> {
```
clintfred: This is a nice change! :tada:

src/policy.rs
```diff
@@ -182,9 +182,9 @@ fn validate_simple_policy_field_value(field_id: &str, field_type: &str) -> Resul
     }
 }
 #[cfg(test)]
-pub(crate) mod test {
+mod tests {
```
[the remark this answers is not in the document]
giarc3: Yes, the convention is `tests`. We had about a 50-50 split of how we were naming them, so I wanted to make them consistent.

src/blocking.rs
giarc3: I was planning to but @coltfred preferred them all together. I'm fine with doing it either way.

src/blocking.rs
giarc3: I briefly looked into an alternative but couldn't find anything. I'm definitely open to alternatives, but I'd need help getting started.

.travis_scripts/cross-test.sh
```diff
@@ -5,13 +5,13 @@ set -x
 
 # If no special target provided, use default cargo arch for building and run all unit tests
 if [ -z "$TARGET" ]; then
-    IRONCORE_ENV=dev cargo t --verbose
+    IRONCORE_ENV=dev cargo t --features "blocking" --verbose
```
ernieturner: Just an optics suggestion, but would it be better to name this "synchronous" instead of "blocking"? The latter seems like it has a more negative connotation for consumers of this SDK.

giarc3: I'm open to naming it something like "synchronous" or "sync". We just chose this name because it matches the flag in `reqwest`, so we sort of accepted it like a standard.

clintfred: I am the one that suggested patterning after reqwest with the blocking terminology. It's certainly open for discussion, but I don't find blocking more negative then sync.

src/blocking.rs
clintfred: In general, when you run out of options to reduce duplication in Rust, procedural macros are the way to go. We leverage the "function-like" variant fairly heavily in both gridiron and recrypt. Here you might be able to use the "attribute" variant here. 

https://doc.rust-lang.org/reference/procedural-macros.html

giarc3: As we decided, I removed the traits, put the functions into `impl BlockingIronOxide` and added hyperlinks to the async docs for all the functions.

src/blocking.rs
```diff
@@ -0,0 +1,325 @@
...
+
+    /// Rotate the private key of the calling user and all groups they are an administrator of where needs_rotation is true.
+    /// Note that this function has the potential to take much longer than other functions, as rotation will be done
+    /// individually on each user/group. If rotation is only needed for a specific group, it is strongly recommended
+    /// to call [user_rotate_private_key()](user\/trait.UserOps.html#tymethod.user_rotate_private_key) or
+    /// [group_rotate_private_key()](group\/trait.GroupOps.html#tymethod.group_rotate_private_key) instead.
+    /// # Arguments
+    /// - `rotations` - PrivateKeyRotationCheckResult that holds all users and groups to be rotated
+    /// - `password` - Password to unlock the current user's user master key
+    pub fn rotate_all(
```
clintfred: Should this link to IronOxide's `rotate_all` for docs?

giarc3: Yes, it should. Fixed.
I have checked that the links work, and I think they look reasonable. 
Example:
image

PR #107, opened 2020-01-14
Add DeviceAddResult

giarc3:
- Added `DeviceAddResult`, the new result of `generate_new_device()`
    - This contains the same fields as `DeviceContext` with the addition of `device_id`, `name`, `created`, and `updated`.
- Added `From<DeviceAddResult> for DeviceContext` to give a `DeviceContext` for functions like `initialize()`
- Updated `DeviceAddResponse` to include new fields

### DIFF
--- a/src/internal/mod.rs
+++ b/src/internal/mod.rs
@@ -372,28 +372,6 @@ pub struct DeviceAddResult {
 }
 
 impl DeviceAddResult {
-    fn new(
-        account_id: UserId,
-        segment_id: usize,
-        device_private_key: PrivateKey,
-        signing_private_key: DeviceSigningKeyPair,
-        device_id: DeviceId,
-        name: Option<DeviceName>,
-        created: DateTime<Utc>,
-        updated: DateTime<Utc>,
-    ) -> DeviceAddResult {
-        DeviceAddResult {
-            account_id,
-            segment_id,
-            device_private_key,
-            signing_private_key,
-            device_id,
-            name,
-            created,
-            updated,
-        }
-    }
-
     pub fn account_id(&self) -> &UserId {
         &self.account_id
     }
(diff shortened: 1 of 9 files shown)


src/internal/mod.rs
```diff
@@ -345,6 +345,71 @@ impl DeviceContext {
...
+        DeviceContext::new(
+            dar.account_id,
+            dar.segment_id,
+            dar.device_private_key,
+            dar.signing_private_key,
+        )
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
clintfred: We need `PartialEq` here. Maybe we should add equality unit tests for all *Result types to ensure that we are providing equality.

I'm also wondering why we are providing `Serialize` and `Deseriaize`. Do we normally provide that? I made a comment on https://github.com/IronCoreLabs/ironoxide/issues/58#issuecomment-574294352 about potentially adding those to all *Result types,  but unless there's a reason for these here, we should probably remove them.

clintfred:
Remember to add a changelog entry!

tests/group_ops.rs
```diff
@@ -165,7 +165,7 @@ fn rotate_all() -> Result<(), IronOxideErr> {
         USER_PASSWORD,
         &Default::default(),
     )?;
-    let creator_sdk = ironoxide::initialize(&device)?;
+    let creator_sdk = ironoxide::initialize(&device.clone().into())?;
```
BobWall23: Why is it necessary to `clone()` here before the `.into()`, when the call below to `initialize_check_rotation()` doesn't need a clone?

giarc3: Good catch.
It was because I was calling the `.into()` twice, which needed ownership of the device, so I had to clone it. I've fixed it now so that `device` has the `.into()` called on it when it's created, so then I can pass references to `initialize()` without needing to clone.

src/internal/mod.rs
giarc3: We can't add `PartialEq` because we don't have it on the device's `PrivateKey`. I've added a comment explaining, and removed the `Serialize` and `Deserialize`

Issue #105, opened 2020-01-10
Empty bytes does not round trip


clintfred:
We allow encrypting an empty array, but decrypting fails with an `AesEncryptedDocSizeError`.

The right thing is probably to error on encrypt, but I'm not 100% sure of that. 

I have created a integration test that shows this on [this branch](https://github.com/IronCoreLabs/ironoxide/tree/bug-encrypt-empty-bytes)

BobWall23:
It should be possible to encrypt an empty string - that shouldn't be an error. Output of encrypt should be an IV and a tag (28 total bytes, I think?)

clintfred:
Encrypt was left alone and the decrypt restriction was relaxed.

PR #108, opened 2020-01-23
[#105] Fixes decryption of data with size 0 bytes and 1 byte

clintfred:
See #105 

Decryption had a check to explicitly disallow "empty" encrypted data. The check was also mistakenly disallowing 1 byte. Both restrictions were relaxed and tests introduced.

### DIFF
--- a/src/crypto/aes.rs
+++ b/src/crypto/aes.rs
@@ -98,8 +98,8 @@ impl TryFrom<&[u8]> for AesEncryptedValue {
     type Error = IronOxideErr;
 
     fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
-        //AES encrypted values should be as long as the IV, GCM auth tag, and at least 1 encrypted byte
-        if bytes.len() <= (AES_IV_LEN + AES_GCM_TAG_LEN + 1) {
+        // AES encrypted values should be as long as the IV, GCM auth tag
+        if bytes.len() < (AES_IV_LEN + AES_GCM_TAG_LEN) {
             Err(IronOxideErr::AesEncryptedDocSizeError)
         } else {
             let mut iv: [u8; AES_IV_LEN] = [0u8; AES_IV_LEN];
@@ -246,6 +246,7 @@ pub fn decrypt(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::convert::TryInto;
     use std::sync::Arc;
 
     #[test]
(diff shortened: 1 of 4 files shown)


src/internal/document_api/mod.rs
```diff
@@ -958,7 +958,10 @@ pub async fn decrypt_document<CR: rand::CryptoRng + rand::RngCore>(
     )?;
 
     aes::decrypt(&mut enc_doc, *sym_key.bytes())
-        .map_err(|e| e.into())
+        .map_err(|e| {
+            println!("map_err: {}", &e);
```
ernieturner: Remove?

src/crypto/aes.rs
```diff
@@ -302,6 +303,33 @@ mod tests {
         assert_eq!(*decrypted_plaintext, plaintext[..]);
     }
 
+    // Even very small documents of 0 and 1 bytes should roundtrip between AesEncryptedValue and bytes
+    #[test]
+    fn test_roundtrip_aesencryptedvalue_zero_one_bytes() -> Result<(), IronOxideErr> {
```
ernieturner: This test name seems like it could be improved or the test changed because you're not really doing a "roundtrip" test right? I would expect a roundtrip to do an encrypt and decrypt and assert you got the same value out.

tests/document_ops.rs
```diff
@@ -21,6 +21,17 @@ extern crate galvanic_assert;
...
 
+#[test]
+fn doc_roundtrip_empty_data() -> Result<(), IronOxideErr> {
+    let sdk = initialize_sdk()?;
+    let doc = [0u8; 0];
+
+    let doc_result = sdk.document_encrypt(&doc, &Default::default())?;
+    let decrypted_result = sdk.document_decrypt(doc_result.encrypted_data())?;
+
+    Ok(assert_eq!(&doc, decrypted_result.decrypted_data()))
```
giarc3: Is there a reason you put the assertion inside of the `Ok` for this test, but not for the other new test you added?
I think we normally just leave the `Ok(())` by itself at the end, though I guess this way is more concise.

src/internal/document_api/mod.rs
clintfred: Yep. Thanks.

tests/document_ops.rs
clintfred: I did it as a single statement because there was only a single assertion at the end of the test. 
I tend to do it this way in simple cases for brevity, but either would be correct in my mind.

src/crypto/aes.rs
clintfred: My intention is that the `AesEncryptedValue` is roundtripping to/from a bytes representation. This is narrower than a full AES encrypt/decrypt. I only included the encrypt step to avoid referencing a the `AES_IV_LEN` and `AES_GCM_TAG_LEN` consts. That was probably a bad idea. I have reworked this test to remove encrypt so hopefully it's more clear what's being tested.

src/internal/document_api/mod.rs
```diff
@@ -957,15 +957,17 @@ pub async fn decrypt_document<CR: rand::CryptoRng + rand::RngCore>(
...
+    Ok(
+        aes::decrypt(&mut enc_doc, *sym_key.bytes()).map(move |decrypted_doc| {
+            DocumentDecryptResult {
+                id: doc_meta.0.id,
+                name: doc_meta.0.name,
+                created: doc_meta.0.created,
+                updated: doc_meta.0.updated,
+                decrypted_data: decrypted_doc.to_vec(),
+            }
+        })?,
```
ernieturner: Just curious, is this a more Rust-preferred way of doing this?

clintfred: I would say yes, so long as you are just calling `into()` on an error and not inspecting the error at all.

Issue #86, opened 2019-12-03
Support Caching of Policy Evaluation


clintfred:
As an ironoxide user
I want SDK functions that require a policy to have the option of caching the result of the evaluated policy
so that round-trips to the server can be reduced and overall performance can be improved.

There are other places where a similar caching mechanism can be used. Policy is the first of these efforts.

AC:
* [ ] Design of a generic, configurable caching layer is complete. This layer should be capable of handling more than just policies.

PR #111, opened 2020-01-30
[#86] Adds policy caching

clintfred:
see #86 

Adds simple policy caching for straightforward use cases where some number of policies are being evaluated repeatedly. Caching will greatly improve the performance of these usecases as it will save a roundtrip to the webservice per encryption.


src/internal/document_api/mod.rs
```diff
@@ -632,6 +637,39 @@ async fn resolve_keys_for_grants(
...
+) -> Result<(Vec<DocAccessEditErr>, Vec<WithKey<UserOrGroup>>), IronOxideErr>
+where
+    F: Future<Output = Result<PolicyResponse, IronOxideErr>>,
+{
+    // if there's a value in the cache, use it
+    if let Some(cached_policy) = policy_cache.get(grant) {
+        Ok((vec![], cached_policy.clone()))
+    } else {
+        // otherwise query the webservice and cache the result if there are no errors
+        match get_policy_f.await {
```
giarc3: Because you're just passing along the error in the `Err` case, this would be cleaned to me as a `map()` instead of a `match`.

src/lib.rs
```diff
@@ -212,19 +272,22 @@ impl PrivateKeyRotationCheckResult {
 
 /// Initialize the IronOxide SDK with a device. Verifies that the provided user/segment exists and the provided device
 /// keys are valid and exist for the provided account. If successful returns an instance of the IronOxide SDK
-pub fn initialize(device_context: &DeviceContext) -> Result<IronOxide> {
+pub fn initialize(device_context: &DeviceContext, config: &IronOxideConfig) -> Result<IronOxide> {
```
giarc3: Don't forget to add the default config to the call in `ironoxide_bench.rs`

src/internal/document_api/mod.rs
clintfred: Absolutely.

tests/document_ops.rs
```diff
@@ -217,7 +217,7 @@ fn doc_create_with_policy_grants() -> Result<(), IronOxideErr> {
         },])
     );
 
-    //finally send an empty policy
+    //finally send an empty policy. This will evaluate cleanly and will thus be cached.
```
coltfred: I don't see a test that verifies that we don't cache unclean evaluations?

src/internal/document_api/mod.rs
```diff
@@ -632,6 +637,36 @@ async fn resolve_keys_for_grants(
...
+    if let Some(cached_policy) = policy_cache.get(grant) {
+        Ok((vec![], cached_policy.clone()))
+    } else {
+        // otherwise query the webservice and cache the result if there are no errors
+        get_policy_f.await.map(|policy_resp| {
+            let (errs, public_keys) = process_policy(&policy_resp);
+            if errs.is_empty() {
+                //if the cache has grown too large, clear it prior to adding new entries
+                if policy_cache.len() >= config.max_entries {
+                    policy_cache.clear()
```
coltfred: I'm a bit sad that this is the invalidation strat we're using. Even removing some arbitrary entry would be better, wouldn't it? Also, you didn't check to see if the if the insert would be actually replacing an entry (concurrent requests).

src/lib.rs
```diff
@@ -93,17 +98,72 @@ use vec1::Vec1;
...
+        }
+    }
+
+    /// Policy evaluation caching config. Lifetime of the cache is the lifetime of the `IronOxide` struct.
+    ///
+    /// Since policies are evaluated by the webservice, caching the result can greatly speed
+    /// up encrypting a document with a [PolicyGrant](../policy/struct.PolicyGrant.html). There is no expiration of the cache, so
+    /// if you want to clear it at runtime, call [IronOxide::clear_policy_cache()](../struct.IronOxide.html#method.clear_policy_cache).
+    #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+    pub struct PolicyCachingConfig {
```
coltfred: I guess we make this now because there will be more configs later? Other option is flattening it into the IronoxideConfig.

src/internal/document_api/mod.rs
```diff
@@ -632,6 +637,36 @@ async fn resolve_keys_for_grants(
     ))
 }
 
+/// Get a cached policy or run the given Future to get the evaluated policy from the webservice.
+/// Policies that evaluate cleanly with no invalid users or groups are cached for future use.
+async fn get_cached_policy_or<F>(
```
coltfred: I love the testibility of this function. So nice!

src/internal/document_api/mod.rs
clintfred: I agree that it's pretty simple minded. All we're really trying to guarantee here is that the cache can't grow without bound for a long running process. I feel like anything smarter probably should include time or usage stats. 

I did consider doing what you suggest and just throwing out a single, random entry, which seems reasonable, but then we need to introduce some machinery to decide _which_ (random) entry to throw out and it just didn't seem worth it.

src/lib.rs
clintfred: Yes. That was my thought. There will be more configs for invalidation.

tests/document_ops.rs
clintfred: This behavior is exhibited by the integration tests, but it would be easy to add as a unit test as well. I'll do this.

Issue #19, opened 2019-07-16
Add configurable timeout to SDK methods


clintfred:
As an SDK user
I want to be able to set timeout on SDK calls
so that I can configure the maximum time I'm willing to wait and handle a timeout in a reasonable manner.

--

Likely, we would make this a global timeout for all operations.

ernieturner:
Discussion needs to be had about whether this timeout applies to entire operation or just for HTTP requests.

clintfred:
Now that the API is built on top of `async fn`, it is straightforward to timeout a top-level SDK call using tokio's [timeout facility](https://docs.rs/tokio/0.2.11/tokio/time/fn.timeout.html). Since the blocking API is built on top of the async one, it would be easy to pass though in that case as well. If this solves most use cases I think this would be a good place to start. It doesn't preclude us from adding network timeouts separately at a later time.

PR #114, opened 2020-02-05
[#19] Adds configurable timeouts to SDK methods

clintfred:
See #19 

* Adds timeouts to all public API methods. Most timeouts use a top-level config set in `IronOxideConfig`. Some special cases allow for passing an optional timeout directly (`rotate_all`, `user_create`, `user_verify`, `generate_new_device`). Timeouts apply to both `IronOxide` and `BlockingIronOxide`
* Configs can now be set on `BlockingIronOxide`. Before, defaults were always used.
* Trying out a "open" struct for all config objects to allow for easier construction and access. If successful, we could use this idea in more in the SDK's public interface.
* Adds dependency on `tokio/rt-threaded` feature flag

### DIFF
--- a/src/blocking.rs
+++ b/src/blocking.rs
@@ -7,7 +7,6 @@
 //!
 //! # Optional
 //! This requires the optional `blocking` feature to be enabled.
-use crate::config::IronOxideConfig;
 pub use crate::internal::{
     document_api::{
         AssociationType, DocAccessEditErr, DocumentAccessResult, DocumentDecryptResult,
@@ -25,6 +24,7 @@ pub use crate::internal::{
     },
 };
 use crate::{
+    config::IronOxideConfig,
     document::{advanced::DocumentAdvancedOps, DocumentEncryptOpts, DocumentOps},
     group::{GroupCreateOpts, GroupOps},
     user::{DeviceCreateOpts, UserCreateOpts, UserOps},
--- a/src/blocking.rs
+++ b/src/blocking.rs
@@ -55,12 +55,13 @@ impl BlockingIronOxide {
         &self,
         rotations: &PrivateKeyRotationCheckResult,
         password: &str,
+        timeout: Option<std::time::Duration>,
     ) -> Result<(
         Option<UserUpdatePrivateKeyResult>,
         Option<Vec<GroupUpdatePrivateKeyResult>>,
     )> {
         self.runtime
-            .enter(|| block_on(self.ironoxide.rotate_all(rotations, password)))
+            .enter(|| block_on(self.ironoxide.rotate_all(rotations, password, timeout)))
     }
 
     /// See [ironoxide::document::DocumentOps::document_list()](trait.DocumentOps.html#tymethod.document_list)
--- a/src/blocking.rs
+++ b/src/blocking.rs
@@ -35,6 +35,7 @@ use crate::{
 use futures::executor::block_on;
 use std::collections::HashMap;
 
+#[derive(Debug)]
 /// Struct that is used to make authenticated requests to the IronCore API. Instantiated with the details
 /// of an account's various ids, device, and signing keys. Once instantiated all operations will be
 /// performed in the context of the account provided. Identical to IronOxide but also contains a Runtime.
--- a/src/blocking.rs
+++ b/src/blocking.rs
@@ -7,6 +7,7 @@
 //!
 //! # Optional
 //! This requires the optional `blocking` feature to be enabled.
+use crate::config::IronOxideConfig;
 pub use crate::internal::{
     document_api::{
         AssociationType, DocAccessEditErr, DocumentAccessResult, DocumentDecryptResult,
@@ -285,9 +286,12 @@ fn create_runtime() -> tokio::runtime::Runtime {
 
 /// Initialize the BlockingIronOxide SDK with a device. Verifies that the provided user/segment exists and the provided device
 /// keys are valid and exist for the provided account. If successful, returns instance of the BlockingIronOxide SDK.
-pub fn initialize(device_context: &DeviceContext) -> Result<BlockingIronOxide> {
+pub fn initialize(
+    device_context: &DeviceContext,
+    config: &IronOxideConfig,
+) -> Result<BlockingIronOxide> {
     let rt = create_runtime();
-    let maybe_io = rt.enter(|| block_on(crate::initialize(device_context, &Default::default())));
+    let maybe_io = rt.enter(|| block_on(crate::initialize(device_context, config)));
     maybe_io.map(|io| BlockingIronOxide {
         ironoxide: io,
         runtime: rt,
@@ -299,14 +303,11 @@ pub fn initialize(device_context: &DeviceContext) -> Result<BlockingIronOxide> {
 /// for private key rotation.
 pub fn initialize_check_rotation(
     device_context: &DeviceContext,
+    config: &IronOxideConfig,
 ) -> Result<InitAndRotationCheck<BlockingIronOxide>> {
     let rt = create_runtime();
-    let maybe_init = rt.enter(|| {
-        block_on(crate::initialize_check_rotation(
-            device_context,
-            &Default::default(),
-        ))
-    });
+    let maybe_init =
+        rt.enter(|| block_on(crate::initialize_check_rotation(device_context, config)));
     maybe_init.map(|init| match init {
         NoRotationNeeded(io) => NoRotationNeeded(BlockingIronOxide {
             ironoxide: io,
--- a/src/blocking.rs
+++ b/src/blocking.rs
@@ -230,7 +230,14 @@ impl BlockingIronOxide {
         user_create_opts: &UserCreateOpts,
     ) -> Result<UserCreateResult> {
         let rt = create_runtime();
-        rt.enter(|| block_on(IronOxide::user_create(jwt, password, user_create_opts)))
+        rt.enter(|| {
+            block_on(IronOxide::user_create(
+                jwt,
+                password,
+                user_create_opts,
+                None,
+            ))
+        })
     }
     /// See [ironoxide::user::UserOps::user_list_devices()](trait.UserOps.html#tymethod.user_list_devices)
     pub fn user_list_devices(&self) -> Result<UserDeviceListResult> {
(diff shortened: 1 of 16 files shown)


tests/blocking_ops.rs
```diff
@@ -86,4 +94,88 @@ mod integration_tests {
...
+                &err_result,
+                has_structure!(IronOxideErr::OperationTimedOut {
+                    operation: eq(SdkOperation::RotateAll),
+                    duration: eq(*duration)
+                })
+            );
+            Ok(())
+        } else {
+            // error type here is bogus, but will cause the test to fail
+            Err(IronOxideErr::InitializeError)
```
giarc3: This could be replaced with something a bit more telling like `panic!("Rotation should have been required")` and not require a bogus error. Also I don't see the benefit of `if let` here compared to a `match`, but I know that's a discussion that several of us have been having.

tests/common/mod.rs
```diff
@@ -169,9 +179,13 @@ pub async fn init_sdk_get_init_result(
         users_device_private_key_bytes.try_into().unwrap(),
         users_signing_keys_bytes.try_into().unwrap(),
     );
+    let config = IronOxideConfig {
```
giarc3: Is there a reason this was specified and not just set to the default?

tests/common/mod.rs
```diff
@@ -106,20 +108,26 @@ pub fn gen_jwt(account_id: Option<&str>) -> (String, String) {
...
         USER_PASSWORD,
         &UserCreateOpts::new(false),
+        None,
     )
     .await?;
     let device = IronOxide::generate_new_device(
         &gen_jwt(Some(account_id.id())).0,
         USER_PASSWORD,
         &Default::default(),
+        None,
```
giarc3: Should this and line 120 use `config.sdk_operation_timeout` to keep the timeouts consistent?

src/blocking.rs
```diff
@@ -228,7 +231,14 @@ impl BlockingIronOxide {
         user_create_opts: &UserCreateOpts,
     ) -> Result<UserCreateResult> {
         let rt = create_runtime();
-        rt.enter(|| block_on(IronOxide::user_create(jwt, password, user_create_opts)))
+        rt.enter(|| {
+            block_on(IronOxide::user_create(
+                jwt,
+                password,
+                user_create_opts,
+                None,
```
giarc3: Shouldn't this function take a `timeout` like `generate_new_device()` and `user_verify()`?

tests/user_ops.rs
```diff
@@ -1,17 +1,22 @@
 mod common;
 
 use common::{create_id_all_classes, gen_jwt, initialize_sdk};
+use galvanic_assert::{
+    matchers::{collection::contains_in_any_order, eq},
+    *,
+};
+use ironoxide::config::IronOxideConfig;
```
giarc3: This is unused

tests/user_ops.rs
```diff
@@ -1,17 +1,22 @@
 mod common;
 
 use common::{create_id_all_classes, gen_jwt, initialize_sdk};
+use galvanic_assert::{
+    matchers::{collection::contains_in_any_order, eq},
```
giarc3: `collection::contains_in_any_order` is unused

src/internal/mod.rs
```diff
@@ -17,12 +18,15 @@ use recrypt::api::{
 };
 use regex::Regex;
 use reqwest::Method;
+use serde::export::fmt::Error;
+use serde::export::Formatter;
```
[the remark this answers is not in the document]
clintfred: I think the serde is probably a reexport. Changed back to std.

src/internal/mod.rs
```diff
@@ -838,11 +918,43 @@ fn gen_plaintext_and_aug_with_retry<R: CryptoOps>(
...
+/// If a timeout limit is reached, the result will be an IronOxideErr::OperationTimedOut.
+/// If no timeout is specified, or if the operation finishes before the timeout, the
+/// result is the result of the sdk operation.
+pub async fn run_maybe_timed_sdk_op<F>(
+    f: F,
+    timeout: Option<std::time::Duration>,
+    op: SdkOperation,
+) -> Result<F::Output, IronOxideErr>
+where
+    F: Future,
```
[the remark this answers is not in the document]
clintfred: I generally don't care either way. For one simple thing, using the non-where style makes sense.

src/internal/mod.rs
```diff
@@ -838,11 +918,43 @@ fn gen_plaintext_and_aug_with_retry<R: CryptoOps>(
...
+{
+    use futures::future::TryFutureExt;
+    let result = match timeout {
+        Some(d) => {
+            tokio::time::timeout(d, f)
+                .map_err(|_: Elapsed| IronOxideErr::OperationTimedOut {
```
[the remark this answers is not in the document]
clintfred: There was some confusion earlier if this was mapping all IronOxideErrs to OperationTimedOut, so I'm going to leave it.

src/lib.rs
```diff
@@ -124,26 +125,16 @@ pub mod config {
...
-        pub fn new(max_entries: usize) -> PolicyCachingConfig {
-            PolicyCachingConfig { max_entries }
-        }
-
-        pub fn max_entries(&self) -> usize {
-            self.max_entries
-        }
+        /// maximum number of policy evaluations that will be cached by the SDK.
+        /// If the maximum number is exceeded, the cache will be cleared prior to storing the next entry
+        pub max_entries: usize,
```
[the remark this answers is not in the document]
clintfred: Data structure sizes are a good candidate for `usize`, I think. The underlying hashmap also uses `usize`.

tests/blocking_ops.rs
```diff
@@ -86,4 +94,88 @@ mod integration_tests {
...
+            &gen_jwt(Some(account_id.id())).0,
+            USER_PASSWORD,
+            &Default::default(),
+            None,
+        )?
+        .into();
+
+        // set a timeout that is unreasonably small
+        let duration = Duration::from_millis(5);
+        let config = IronOxideConfig {
```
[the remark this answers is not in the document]
clintfred: It is used on 122.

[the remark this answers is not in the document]
giarc3: Yep, I commented on the wrong one. The one on line 155 is unused

tests/blocking_ops.rs
clintfred: I think there are usecases for both, and there are some where they are both equally valid. In this case I only want to handle one case which is great for if let. I do like the `panic` suggestion.

tests/common/mod.rs
clintfred: Mostly because I don't know what the default will end up being. ;)

tests/common/mod.rs
clintfred: I agreed with this at first, but after looking at it more, the name says `init_sdk_with_config`, so I think the other timeouts for `user_create` and `generate_new_device` are separate concerns. I came to this conclusion after running a test that was expecting `initialize` to timeout after passing the config value, but once I reused the timeout from the config, `user_create` started timing out first.

src/internal/mod.rs
```diff
@@ -72,6 +76,79 @@ pub enum RequestErrorCode {
     PolicyGet,
 }
 
+/// Public SDK operations
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Copy)]
+pub enum SdkOperation {
```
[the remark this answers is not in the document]
clintfred: Enumerating all the strings as enum variants  is a good thing, IMO. I think the bigger issue is the match you commented on below.

src/blocking.rs
clintfred: Yes. Great catch.

src/internal/mod.rs
```diff
@@ -72,6 +76,79 @@ pub enum RequestErrorCode {
...
+    GroupAddMembers,
+    GroupRemoveMembers,
+    GroupAddAdmins,
+    GroupRemoveAdmins,
+    GroupRotatePrivateKey,
+}
+
+impl std::fmt::Display for SdkOperation {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), Error> {
+        let name = match self {
```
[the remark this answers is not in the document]
clintfred: I couldn't make the code you posted or anything similar do anything reasonable. I ended up using `Debug` for enum to write `Display` to get a decent string representation out of the enum. We could copy the code generated by `Debug` or do something else too...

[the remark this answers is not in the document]
clintfred: BTW: The generated `Debug` impl does something very similar to what the original implementation here did: pattern match and generate strings. If we had a lot of these, we could write our own macro, I guess...

src/document/advanced.rs
```diff
@@ -70,33 +72,41 @@ impl DocumentAdvancedOps for crate::IronOxide {
...
-            &self.user_master_pub_key,
-            &self.rng,
-            data,
-            encrypt_opts.id.clone(),
-            grant_to_author,
-            &explicit_users,
-            &explicit_groups,
-            policy_grants,
+        run_maybe_timed_sdk_op(
+            internal::document_api::encrypted_document_unmanaged(
```
giarc3: Not part of the work you're doing, but I'm surprised that `encrypted_document_unmanaged()` is that function's name

clintfred: Bad name, indeed. It's internal, so easily changed. New name will be `encrypt_document_unmanaged`

src/internal/mod.rs
```diff
@@ -838,11 +885,40 @@ fn gen_plaintext_and_aug_with_retry<R: CryptoOps>(
     aug_private_key().or_else(|_| aug_private_key())
 }
 
+/// Runs a future with a timeout or just runs the future, depending on if a timeout is specified.
+///
+/// If a timeout limit is reached, the result will be an IronOxideErr::OperationTimedOut.
+/// If no timeout is specified, or if the operation finishes before the timeout, the
+/// result is the result of the sdk operation.
+pub async fn run_maybe_timed_sdk_op<F: Future>(
```
coltfred: This doesn't run anything, it just adds timeout to it. Name suggestion `add_future_timeout` or `add_timeout` and let the fact that it takes a future do the talking.

src/blocking.rs
```diff
@@ -34,6 +35,7 @@ use crate::{
 use futures::executor::block_on;
 use std::collections::HashMap;
 
+#[derive(Debug)]
```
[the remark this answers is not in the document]
clintfred: Yep. Not sure what happened there.

src/internal/mod.rs
clintfred: Went with `add_optional_timeout` -- seem better?

PR #117, opened 2020-02-19
Various cleanups with cargo clippy

giarc3:
If there are any lints you disagree with, I can remove those changes.
I don't think I made any breaking changes, but could be mistaken.

### DIFF
--- a/src/crypto/aes.rs
+++ b/src/crypto/aes.rs
@@ -7,7 +7,7 @@ use crate::internal::{take_lock, IronOxideErr};
 use std::{convert::TryFrom, ops::DerefMut, sync::Mutex};
 
 //There is no way this can fail. Value is most definitely not less than one.
-const PBKDF2_ITERATIONS: NonZeroU32 = unsafe { NonZeroU32::new_unchecked(250000) };
+const PBKDF2_ITERATIONS: NonZeroU32 = unsafe { NonZeroU32::new_unchecked(250_000) };
 const PBKDF2_SALT_LEN: usize = 32;
 const AES_GCM_TAG_LEN: usize = 16;
 const AES_IV_LEN: usize = 12;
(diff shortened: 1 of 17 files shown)


src/crypto/aes.rs
```diff
@@ -7,7 +7,7 @@ use crate::internal::{take_lock, IronOxideErr};
 use std::{convert::TryFrom, ops::DerefMut, sync::Mutex};
 
 //There is no way this can fail. Value is most definitely not less than one.
-const PBKDF2_ITERATIONS: NonZeroU32 = unsafe { NonZeroU32::new_unchecked(250000) };
+const PBKDF2_ITERATIONS: NonZeroU32 = unsafe { NonZeroU32::new_unchecked(250_000) };
```
clintfred: We'd discussed this formatting before and decided against it. I'm fine with it, however, if others are.

src/internal/user_api/mod.rs
```diff
@@ -56,7 +56,7 @@ impl TryFrom<u64> for DeviceId {
         //Validate the range of the device ID to always be positive, but also be
         //less than i64 (i.e. no high bit set) for compatibility with other
         //languages (i.e. Java)
-        if device_id < 1 || device_id > (std::i64::MAX as u64) {
+        if device_id < 1 || device_id > (i64::max_value() as u64) {
```
clintfred: This surprises me a bit, but I'm guessing that `max_value()` is a const fn so there is no difference at runtime. Has `MAX` been deprecated, or why is this better?

src/crypto/aes.rs
giarc3: I've talked to @coltfred about it in the past, and our consensus was to use it sparingly only when it makes sense. So here I think it improves readability, but numbers in gridiron are mostly incomprehensible anyway, so it would only make it more cluttered.

src/internal/user_api/mod.rs
giarc3: Quick googling suggests that they're equivalent, so I don't think either is better. Not sure why there's a lint for it, but it shouldn't hurt? I can change this one back if you prefer.

clintfred: No. No problem, I was just wondering. This is fine by me.

clintfred:
Still wondering about cargo clippy integration for Travis...

giarc3:
> Still wondering about cargo clippy integration for Travis...

In my mind, putting a clippy requirement into Travis will be more of an annoyance than it's worth. I don't run clippy regularly, and I don't think others do either, so I'm guessing it will be another roadblock to merging that slows us down for a small benefit. And as we've all become more familiar with Rust, I think we've naturally become better at following its guidelines anyway.
How do you feel about it?

clintfred:
I was under the impression that this PR was giving us a clean `clippy` report, but that doesn't seem to be the case. We just used it as a guide, and that's fine, and in that case integrating it into Travis would, indeed, just be noise. Sorry for the confusion.

PR #118, opened 2020-02-20
Blind index

coltfred:
TODO
- [x] Tests
- [x] Clarify use of the words index and indexed terms

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ vec1 = "~1.4.0"
 # ironoxide requires rt-threaded at runtime, but not at compile time
 tokio = {version = "~0.2.0", features=["time", "rt-threaded"]}
 dashmap = "3.4.0"
-ironcore-search-helpers = { git = "https://github.com/IronCoreLabs/search-helpers", rev = "cf9d8bd" }
+ironcore-search-helpers = { git = "https://github.com/IronCoreLabs/search-helpers", rev = "cf9d8bd", optional=true}
 
 [dev-dependencies]
 frank_jwt = "~3.1.2"
@@ -58,6 +58,7 @@ all-features = true
 
 [features]
 blocking = ["tokio/rt-threaded"]
+beta = ["ironcore-search-helpers"]
 
 [profile.dev]
 opt-level = 2
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ vec1 = "~1.4.0"
 # ironoxide requires rt-threaded at runtime, but not at compile time
 tokio = {version = "~0.2.0", features=["time", "rt-threaded"]}
 dashmap = "3.4.0"
-ironcore-search-helpers = { git = "https://github.com/IronCoreLabs/search-helpers", rev = "cf9d8bd", optional=true}
+ironcore-search-helpers = { git = "https://github.com/IronCoreLabs/search-helpers", rev = "38be96fe6160a739396cf505054a4e0dec6e0aa6", optional=true}
 
 [dev-dependencies]
 frank_jwt = "~3.1.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ vec1 = "~1.4.0"
 # ironoxide requires rt-threaded at runtime, but not at compile time
 tokio = {version = "~0.2.0", features=["time", "rt-threaded"]}
 dashmap = "3.4.0"
+ironcore-search-helpers = { git = "https://github.com/IronCoreLabs/search-helpers", rev = "cf9d8bd" }
 
 [dev-dependencies]
 frank_jwt = "~3.1.2"
--- a/src/search.rs
+++ b/src/search.rs
@@ -140,7 +140,7 @@ impl BlindIndexSearch {
         }
     }
 
-    ///Generate the list of tokens to use to find entries that match the search query, given the specified partition_id.
+    /// Generate the list of tokens to use to find entries that match the search query, given the specified partition_id.
     /// query - The string you want to tokenize and hash
     /// partition_id - An extra string you want to include in every hash, this allows 2 queries with different partition_ids to produce a different set of tokens for the same query
     pub fn tokenize_query(&self, query: &str, partition_id: Option<&str>) -> Result<HashSet<u32>> {
@@ -148,8 +148,11 @@ impl BlindIndexSearch {
             .map_err(|message| IronOxideErr::ValidationError("query".to_string(), message))
     }
 
-    ///Generate the list of tokens to use to find entries that match the search query, given the specified partition_id.
-    /// query - The string you want to tokenize and hash
+    /// Generate the list of tokens to create a search entry for `data`. This function will also return some random values in the HashSet, which will make
+    /// it harder for someone to know what the input was. Because of this, calling this function will not be the same as `tokenize_query`, but `tokenize_query` will always
+    /// return a subset of the values returned by `tokenize_data`.
+    ///
+    /// data - The string you want to tokenize and hash
     /// partition_id - An extra string you want to include in every hash, this allows 2 queries with different partition_ids to produce a different set of tokens for the same data
     pub fn tokenize_data(&self, data: &str, partition_id: Option<&str>) -> Result<HashSet<u32>> {
         generate_hashes_for_string_with_padding(
--- a/src/search.rs
+++ b/src/search.rs
@@ -22,6 +22,9 @@ use std::ops::DerefMut;
 
 use ironcore_search_helpers::generate_hashes_for_string;
 
+#[cfg(feature = "blocking")]
+use crate::blocking::BlockingIronOxide;
+
 ///The required length of the salt.
 const REQUIRED_LEN: usize = 32;
 
@@ -37,6 +40,12 @@ impl EncryptedBlindIndexSalt {
             .document_decrypt_unmanaged(&self.encrypted_salt_bytes[..], &self.encrypted_deks[..])
             .await?;
         decrypted_value.decrypted_data().try_into()
+    }
+
+    #[cfg(feature = "blocking")]
+    pub fn initialize_search_blocking(&self, bio: &BlockingIronOxide) -> Result<BlindIndexSearch> {
+        bio.runtime
+            .enter(|| futures::executor::block_on(self.initialize_search(&bio.ironoxide)))
     }
 }
 
@@ -70,6 +79,7 @@ impl BlindIndexSearchInitialize for IronOxide {
         encrypted_salt.try_into()
     }
 }
+
 #[derive(Debug, PartialEq, Clone, Hash, Eq)]
 pub struct BlindIndexSearch {
     decrypted_salt: [u8; 32],
--- a/src/search.rs
+++ b/src/search.rs
@@ -28,6 +28,7 @@ use crate::blocking::BlockingIronOxide;
 ///The required length of the salt.
 const REQUIRED_LEN: usize = 32;
 
+#[derive(Debug, PartialEq, Clone, Hash, Eq)]
 pub struct EncryptedBlindIndexSalt {
     pub encrypted_deks: Vec<u8>,
     pub encrypted_salt_bytes: Vec<u8>,
--- a/src/search.rs
+++ b/src/search.rs
@@ -25,22 +25,25 @@ pub struct CreatedIndexResult {
 #[async_trait]
 pub trait SimpleSeachInitialize {
     ///Given the encrypted salt and the edeks, decrypt them and give back the IronSimpleSearch object.
-    async fn initialize_search_index(
+    async fn initialize_search_sdk(
         &self,
         encrypted_data: &[u8],
         encrypted_deks: &[u8],
     ) -> Result<IronSimpleSearch>;
     ///Create an index and encrypt it to the provided group_id.
-    ///If you need to index terms immediately, see `create_index_and_initialize` which will return
+    ///If you need to index data immediately, see `create_index_and_initialize_search_sdk` which will return
     ///the IronSimpleSearch for reuse.
     async fn create_index(&self, group_id: &GroupId) -> Result<DocumentEncryptUnmanagedResult>;
     ///Create an index, encrypt it and initialize a IronSimpleSearch for immediate use.
-    async fn create_index_and_initialize(&self, group_id: &GroupId) -> Result<CreatedIndexResult>;
+    async fn create_index_and_initialize_search_sdk(
+        &self,
+        group_id: &GroupId,
+    ) -> Result<CreatedIndexResult>;
 }
 
 #[async_trait]
 impl SimpleSeachInitialize for IronOxide {
-    async fn initialize_search_index(
+    async fn initialize_search_sdk(
         &self,
         encrypted_salt: &[u8],
         encrypted_salt_deks: &[u8],
@@ -51,11 +54,15 @@ impl SimpleSeachInitialize for IronOxide {
         decrypted_value.decrypted_data().try_into()
     }
     async fn create_index(&self, group_id: &GroupId) -> Result<DocumentEncryptUnmanagedResult> {
-        let CreatedIndexResult { encrypted_salt, .. } =
-            self.create_index_and_initialize(group_id).await?;
+        let CreatedIndexResult { encrypted_salt, .. } = self
+            .create_index_and_initialize_search_sdk(group_id)
+            .await?;
         Ok(encrypted_salt)
     }
-    async fn create_index_and_initialize(&self, group_id: &GroupId) -> Result<CreatedIndexResult> {
+    async fn create_index_and_initialize_search_sdk(
+        &self,
+        group_id: &GroupId,
+    ) -> Result<CreatedIndexResult> {
         let salt = {
             let mut mut_salt = [0u8; 32];
             take_lock(&self.rng).deref_mut().fill_bytes(&mut mut_salt);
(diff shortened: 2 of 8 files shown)


src/search.rs
```diff
@@ -0,0 +1,141 @@
...
+    ///Given the encrypted salt and the edeks, decrypt them and give back the IronSimpleSearch object.
+    async fn initialize_search_index(
+        &self,
+        encrypted_data: &[u8],
+        encrypted_deks: &[u8],
+    ) -> Result<IronSimpleSearch>;
+    ///Create an index and encrypt it to the provided group_id.
+    ///If you need to index terms immediately, see `create_index_and_initialize` which will return
+    ///the IronSimpleSearch for reuse.
+    async fn create_index(&self, group_id: &GroupId) -> Result<DocumentEncryptUnmanagedResult>;
```
BobWall23: Shouldn't the `DocumentEncryptUnmanagedResult` that this returns be the input `initialize_search_index`?

src/search.rs
```diff
@@ -0,0 +1,141 @@
...
+            .await?;
+        decrypted_value.decrypted_data().try_into()
+    }
+    async fn create_index(&self, group_id: &GroupId) -> Result<DocumentEncryptUnmanagedResult> {
+        let CreatedIndexResult { encrypted_salt, .. } =
+            self.create_index_and_initialize(group_id).await?;
+        Ok(encrypted_salt)
+    }
+    async fn create_index_and_initialize(&self, group_id: &GroupId) -> Result<CreatedIndexResult> {
+        let salt = {
```
BobWall23: What's the benefit of the lambda function over just inlining?

src/search.rs
```diff
@@ -0,0 +1,141 @@
...
+#[async_trait]
+pub trait SimpleSeachInitialize {
+    ///Given the encrypted salt and the edeks, decrypt them and give back the IronSimpleSearch object.
+    async fn initialize_search_index(
+        &self,
+        encrypted_data: &[u8],
+        encrypted_deks: &[u8],
+    ) -> Result<IronSimpleSearch>;
+    ///Create an index and encrypt it to the provided group_id.
+    ///If you need to index terms immediately, see `create_index_and_initialize` which will return
```
BobWall23:
```suggestion
    ///If you need to index data immediately, see `create_index_and_initialize` which will return
```

src/search.rs
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+}
+
+impl IronSimpleSearch {
+    fn new(decrypted_salt: [u8; 32]) -> IronSimpleSearch {
+        IronSimpleSearch { decrypted_salt }
+    }
+
+    ///Generate the search query to try and find term in a pariticular partition_id.
```
BobWall23:
```suggestion
    ///Generate the list of tokens to use to find entries that match the search query, given the specified partition_id.
```

src/search.rs
```diff
@@ -0,0 +1,141 @@
...
+    pub fn generate_query(&self, term: &str, partition_id: Option<&str>) -> Vec<u32> {
+        generate_hashes_for_string(term, partition_id, &self.decrypted_salt[..])
+    }
+
+    ///Create a new index for the term in partition_id.
```
BobWall23:
```suggestion
    ///Generate the list of search tokens that represent the supplied string.
```

src/search.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+}
+
+impl IronSimpleSearch {
+    fn new(decrypted_salt: [u8; 32]) -> IronSimpleSearch {
+        IronSimpleSearch { decrypted_salt }
+    }
+
+    ///Generate the search query to try and find term in a pariticular partition_id.
+    pub fn generate_query(&self, term: &str, partition_id: Option<&str>) -> Vec<u32> {
```
BobWall23:
```suggestion
    pub fn generate_query(&self, query_string: &str, partition_id: Option<&str>) -> Vec<u32> {
```

src/search.rs
```diff
@@ -0,0 +1,141 @@
...
+        generate_hashes_for_string(term, partition_id, &self.decrypted_salt[..])
```
BobWall23:
```suggestion
        generate_hashes_for_string(query_string, partition_id, &self.decrypted_salt[..])
```

src/search.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    ///Create a new index for the term in partition_id.
+    pub fn generate_index_tokens(&self, term: &str, partition_id: Option<&str>) -> Vec<u32> {
```
BobWall23:
```suggestion
    pub fn generate_index_tokens(&self, string_to_index: &str, partition_id: Option<&str>) -> Vec<u32> {
```

src/search.rs
```diff
@@ -0,0 +1,141 @@
...
+        self.generate_query(term, partition_id)
```
BobWall23:
```suggestion
        self.generate_query(string_to_index, partition_id)
```

src/search.rs
coltfred: The `DocumentEncryptUnmanagedResult` has the people it was encrypted to, errors about the sharing, etc. Maybe I should change the structure to just include the relevant data.

src/search.rs
coltfred: This isn't a lambda, just limited scoping of the mutable variable.

src/search.rs
```diff
@@ -0,0 +1,147 @@
...
+            .await?;
+        let search_sdk = IronSimpleSearch::new(salt);
+
+        Ok(CreatedIndexResult {
+            encrypted_salt,
+            sdk: search_sdk,
+        })
+    }
+}
+#[derive(Debug, PartialEq, Clone)]
```
clintfred: Any reason not to just add `Hash` and `Eq` here too?

src/search.rs
```diff
@@ -0,0 +1,147 @@
...
+///The result of creating a new index as well as initializing an IronSimpleSdk.
+///If you only want to create the index, see create_index.
+pub struct CreatedIndexResult {
+    pub encrypted_salt: DocumentEncryptUnmanagedResult,
+    pub sdk: IronSimpleSearch,
+}
+
+///Trait which gives the ability to create an index.
+#[async_trait]
+pub trait SimpleSeachInitialize {
```
clintfred: Typo: Seach -> Search

src/search.rs
```diff
@@ -0,0 +1,147 @@
...
+    ///Given the encrypted salt and the edeks, decrypt them and give back the IronSimpleSearch object.
+    async fn initialize_search_sdk(
+        &self,
+        encrypted_data: &[u8],
```
clintfred: Should we just call this the `encrypted_salt`?

src/search.rs
```diff
@@ -0,0 +1,147 @@
...
+
+use ironcore_search_helpers::generate_hashes_for_string;
+
+type Result<A> = std::result::Result<A, IronOxideErr>;
+///The required length of the salt.
+const REQUIRED_LEN: usize = 32;
+
+///The result of creating a new index as well as initializing an IronSimpleSdk.
+///If you only want to create the index, see create_index.
+pub struct CreatedIndexResult {
```
clintfred: Suggest: `CreateIndexResult` or `SearchIndexCreateResult` or `IndexCreateResult`

src/search.rs
clintfred: Or maybe give it an alias that better describes its usage here?

src/search.rs
giarc3: Of the suggestions, I would choose `IndexCreateResult`

BobWall23: Shouldn't `create_index` return an `index`?

src/search.rs
coltfred: I changed it to encrypted_salt_bytes to hopefully avoid confusion between the entire "EncryptedSalt" which is the deks + bytes and just the Aes bytes.

src/search.rs
coltfred: No, added.

src/search.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub encrypted_salt: DocumentEncryptUnmanagedResult,
+    pub sdk: IronSimpleSearch,
```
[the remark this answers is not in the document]
coltfred: I think I disagree on this point. I've shifted to a more specific naming around BlindIndex.

src/search.rs
clintfred: What do you think about putting the encrypted_salt in the BlindIndexSearchSdk struct and getting rid of this altogether?

Maybe this would look like getting rid of `create_index` (or not) and changing `create_index_and_initialize_search` to also return `Result<BlindIndexSearchSdk>`

src/search.rs
```diff
@@ -0,0 +1,166 @@
...
+    pub encrypted_salt_bytes: Vec<u8>,
+}
+
+///Trait which gives the ability to create an blind index.
+#[async_trait]
+pub trait BlindIndexSearchInitialize {
+    ///Given the encrypted blind index salt, decrypt it and give back the BlindIndexSearchSdk object.
+    async fn initialize_search(
+        &self,
+        search: &EncryptedBlindIndexSalt,
```
clintfred: rename to `encrypted_salt`?

src/search.rs
```diff
@@ -0,0 +1,166 @@
...
+    ///Create an index, encrypt it and initialize a BlindIndexSearchSdk for immediate use.
+    async fn create_index_and_initialize_search(
+        &self,
+        group_id: &GroupId,
+    ) -> Result<CreatedIndexResult>;
+}
+
+#[async_trait]
+impl BlindIndexSearchInitialize for IronOxide {
+    async fn initialize_search(
```
clintfred: consider rename to `initialize_blind_index_search`?

src/search.rs
```diff
@@ -0,0 +1,166 @@
...
+                encrypted_salt_bytes: r.encrypted_data().to_vec(),
+            }),
+            Some(err) => Err(IronOxideErr::UserOrGroupDoesNotExist(
+                err.user_or_group.clone(),
+            )),
+        }
+    }
+}
+
+impl BlindIndexSearchSdk {
```
clintfred: I think `BlindIndexSearch` is a slightly more succinct name with the same meaning

src/search.rs
```diff
@@ -0,0 +1,166 @@
...
+    fn new(decrypted_salt: [u8; 32]) -> BlindIndexSearchSdk {
+        BlindIndexSearchSdk { decrypted_salt }
+    }
+
+    ///Generate the list of tokens to use to find entries that match the search query, given the specified partition_id.
+    pub fn tokenize_query(&self, query: &str, partition_id: Option<&str>) -> Vec<u32> {
```
clintfred: I don't understand what `partition_id` is. These two functions could use more descriptive docs, perhaps with an example.

coltfred: Added some more text about it.

src/search.rs
coltfred: ok.

src/search.rs
coltfred: ok.

src/search.rs
coltfred: ok.

src/search.rs
```diff
@@ -0,0 +1,177 @@
...
+    ///Generate the list of tokens to use to find entries that match the search query, given the specified partition_id.
+    /// query - The string you want to tokenize and hash
+    /// partition_id - An extra string you want to include in every hash, this allows 2 queries with different partition_ids to produce a different set of
+    pub fn tokenize_query(&self, query: &str, partition_id: Option<&str>) -> HashSet<u32> {
+        generate_hashes_for_string(query, partition_id, &self.decrypted_salt[..])
+    }
+
+    ///Generate the list of tokens to use to find entries that match the search query, given the specified partition_id.
+    /// query - The string you want to tokenize and hash
+    /// partition_id - An extra string you want to include in every hash, this allows 2 queries with different partition_ids to produce a different set of
```
clintfred: I think you forgot to end your

src/search.rs
clintfred: You said "ok", but didn't change it. Intentional?

src/search.rs
clintfred: Discussed offline that this type of construction doesn't make sense as the existence of a `BlindIndexSearch` doesn't necessarily mean that an `EncryptedBlindIndexSalt` exists.

So I'm back to suggesting `IndexCreateResult` as a name here, unless someone has something better. I think what bothers me about the current name is that it starts with a verb.

tests/search_ops.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    #[tokio::test]
+    async fn initialize_blind_index_search() -> Result<(), IronOxideErr> {
+        let sdk = initialize_sdk().await?;
+        let group_create_result = sdk.group_create(&Default::default()).await?;
+        let result = sdk
+            .initialize_blind_index_search(group_create_result.id())
+            .await?;
+        let index_tokens = result.sdk.tokenize_data("hello world", Option::None);
+        assert_eq!(index_tokens.len(), 6); //hel, ell, elo, wor, orl, rld --The numbers are random and aren't worth asserting about.
```
clintfred: and the stability of the token generation itself is already covered by tests in search-helpers, correct?

src/search.rs
coltfred: Sorry. Pushed.

src/search.rs
coltfred: I don't know what

tests/search_ops.rs
coltfred: Yes.

Cargo.toml
```diff
@@ -40,6 +40,7 @@ vec1 = "~1.4.0"
 # ironoxide requires rt-threaded at runtime, but not at compile time
 tokio = {version = "~0.2.0", features=["time", "rt-threaded"]}
 dashmap = "3.4.0"
+ironcore-search-helpers = { git = "https://github.com/IronCoreLabs/search-helpers", rev = "38be96fe6160a739396cf505054a4e0dec6e0aa6", optional=true}
```
ernieturner: Is the plan to still add tags to this repo so we can point to a specific tag for versioning?

src/search.rs
```diff
@@ -0,0 +1,182 @@
...
+    ///Generate the list of tokens to use to find entries that match the search query, given the specified partition_id.
+    /// query - The string you want to tokenize and hash
+    /// partition_id - An extra string you want to include in every hash, this allows 2 queries with different partition_ids to produce a different set of tokens for the same query
+    pub fn tokenize_query(&self, query: &str, partition_id: Option<&str>) -> HashSet<u32> {
+        generate_hashes_for_string(query, partition_id, &self.decrypted_salt[..])
+    }
+
+    ///Generate the list of tokens to use to find entries that match the search query, given the specified partition_id.
+    /// query - The string you want to tokenize and hash
+    /// partition_id - An extra string you want to include in every hash, this allows 2 queries with different partition_ids to produce a different set of tokens for the same data
```
ernieturner: This comment looks like it was just copy/pasted from `tokenize_query` where I assume it serves a different purpose even though the arguments are the same type.

src/search.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub sdk: BlindIndexSearch,
+}
+
+pub struct EncryptedBlindIndexSalt {
+    pub encrypted_deks: Vec<u8>,
+    pub encrypted_salt_bytes: Vec<u8>,
+}
+
+impl EncryptedBlindIndexSalt {
+    //ecrypt the blind index salt and give back the BlindIndexSearch object.
```
giarc3:
```suggestion
    //encrypt the blind index salt and give back the BlindIndexSearch object.
```

src/search.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub async fn initialize_search(&self, ironoxide: &IronOxide) -> Result<BlindIndexSearch> {
+        let decrypted_value = ironoxide
+            .document_decrypt_unmanaged(&self.encrypted_salt_bytes[..], &self.encrypted_deks[..])
+            .await?;
+        decrypted_value.decrypted_data().try_into()
+    }
+}
+
+///Trait which gives the ability to create an blind index.
```
giarc3:
```suggestion
///Trait which gives the ability to create a blind index.
```

src/search.rs
```diff
@@ -0,0 +1,156 @@
...
+
+///The required length of the salt.
+const REQUIRED_LEN: usize = 32;
+
+///The result of creating a new index as well as initializing a BlindIndexSearch.
+///If you only want to create the index, see create_index.
+pub struct BlindIndexCreateResult {
+    pub encrypted_salt: EncryptedBlindIndexSalt,
+    pub sdk: BlindIndexSearch,
+}
```
ernieturner: This looks to be unused now, which matches up with what we discussed in the meeting.

src/search.rs
```diff
@@ -0,0 +1,156 @@
...
+pub struct BlindIndexSearch {
+    decrypted_salt: [u8; 32],
+}
+
+impl TryFrom<&[u8]> for BlindIndexSearch {
+    type Error = IronOxideErr;
+    fn try_from(bytes: &[u8]) -> Result<BlindIndexSearch> {
+        let decrypted_len = bytes.len();
+        if decrypted_len != REQUIRED_LEN {
+            std::result::Result::Err(IronOxideErr::WrongSizeError(
```
giarc3:
```suggestion
            Err(IronOxideErr::WrongSizeError(
```

Cargo.toml
coltfred: Turns out we have to publish it or we can't publish ironoxide. I'll do that.

src/search.rs
coltfred: Removed

src/search.rs
coltfred: That's because they currently do the same thing. I'll change it when it gets updated to do something else.

giarc3: That time has come

coltfred: I changed the comment. Let me know if you think it could be improved.

PR #119, opened 2020-02-28
Add derives for Clone, Debug, Eq, Hash, and PartialEq for public types

giarc3:
This makes sure that Clone, Debug, Eq, Hash, and PartialEq are on all of our public types. 

Exceptions:
- IronOxide/BlockingIronOxide only have Debug, as parts of them like recrypt stop them from having others. 
- build.rs-generated structs like `ironoxide::proto::transform::PublicKey` only have Clone and PartialEq. I looked into finding a way to make it add others with no success. The full list is `EncryptedDekData`, `EncryptedDeks`, `PublicKey`, `UserOrGroup`, and the enum `UserOrGroup_oneof_userOrGroupId`.

TODO:
- [x] Depend on released recrypt-rs
- [x] Changelog

### DIFF
--- a/src/internal/document_api/mod.rs
+++ b/src/internal/document_api/mod.rs
@@ -46,7 +46,7 @@ const HEADER_META_LENGTH_LENGTH: usize = 2;
 const CURRENT_DOCUMENT_ID_VERSION: u8 = 2;
 
 /// Document ID. Unique within the segment. Must match the regex `^[a-zA-Z0-9_.$#|@/:;=+'-]+$`
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub struct DocumentId(pub(crate) String);
 impl DocumentId {
     pub fn id(&self) -> &str {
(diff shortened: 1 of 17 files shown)


src/internal/document_api/mod.rs
```diff
@@ -46,7 +46,7 @@ const HEADER_META_LENGTH_LENGTH: usize = 2;
 const CURRENT_DOCUMENT_ID_VERSION: u8 = 2;
 
 /// Document ID. Unique within the segment. Must match the regex `^[a-zA-Z0-9_.$#|@/:;=+'-]+$`
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
```
clintfred: Looks like you are ordering these consistently, although this one is different....

giarc3: Good catch. Fixed.

clintfred:
I did open https://github.com/rust-itertools/itertools/pull/417 to add `Hash` to `EitherOrBoth`. It won't matter for this PR, probably, but it would let us delete a couple Hash implementations.

clintfred:
That PR was merged.
So we _could_ add the following to our Cargo.toml and remove the `Hash` impls created due to `EitherOrBoth`, or write a ticket and wait for the next release. I scanned through the commits to master and I don't see anything that concerns me, so I would do it, but others may disagree. Their release schedule is certainly sporadic.

```
# itertools 0.8.2++ - remove on next release
itertools = {git = "https://github.com/rust-itertools/itertools", rev = "c83cc48ffe39b96b7c6797b7b31752e840b373bb"}
```

PR #120, opened 2020-03-02
Add clear_policy_cache() to BlockingIronOxide

giarc3:
Add function to BlockingIronOxide to clear policy cache. Simply calls `clear_policy_cache()` on its internal IronOxide.

### DIFF
--- a/src/blocking.rs
+++ b/src/blocking.rs
@@ -50,6 +50,13 @@ impl BlockingIronOxide {
         &self.ironoxide.device
     }
 
+    /// Clears all entries from the policy cache.
+    /// # Returns
+    /// Number of entries cleared from the cache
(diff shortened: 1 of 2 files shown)


src/blocking.rs
```diff
@@ -50,6 +50,13 @@ impl BlockingIronOxide {
         &self.ironoxide.device
     }
 
+    /// Clears all entries from the policy cache.
+    /// # Returns
+    /// Number of entries cleared from the cache
```
coltfred: Maybe we should use a See comment instead. ` /// See [ironoxide::document::DocumentOps::document_list()](trait.DocumentOps.html#tymethod.document_list)` for instance

giarc3: Done

PR #126, opened 2020-03-11
Upgrade to recrypt 0.11, rand 0.7

giarc3:


Cargo.toml
```diff
@@ -56,7 +56,7 @@ protobuf-codegen-pure = "~2.10"
 all-features = true
 
 [features]
-blocking = ["tokio/rt-threaded"]
+blocking = []
```
giarc3: Removed this because we changed it to be always enabled at the dependency level (line 41)

PR #129, opened 2020-03-19
Expose masked error message on SDK initialization

clintfred:


src/internal/document_api/mod.rs
```diff
@@ -1330,7 +1330,7 @@ mod tests {
 
         // let's get the policy again, but if the policy future executes (cache miss) error
         get_cached_policy_or(&config, &policy_grant, &policy_cache, async {
-            Err(IronOxideErr::InitializeError)
+            Err(IronOxideErr::InitializeError("".into()))
```
giarc3: Seems like `"SDK initialization failed. Underlying cause ''"` will be confusing. Maybe `InitializeError` could take `Option<String>` so we can differentiate when we want the generic error message vs. a specific one?

CHANGELOG.md
```diff
@@ -1,10 +1,13 @@
...
+## 0.21.1 (unreleased)
+- [[#129](https://github.com/IronCoreLabs/ironoxide/pull/129)] Improved error message for SDK initialization failure
+
 ## 0.21.0
 
-- [[126](https://github.com/IronCoreLabs/ironoxide/pull/126)]
+- [[#126](https://github.com/IronCoreLabs/ironoxide/pull/126)]
   - Upgrade dependencies (`rand`, `rand_chacha`, `recrypt`)
-- [[118]](https://github.com/IronCoreLabs/ironoxide/pull/118)
+- [[#118](https://github.com/IronCoreLabs/ironoxide/pull/118)]
```
giarc3: Thanks for catching this

CHANGELOG.md
```diff
@@ -1,10 +1,13 @@
 # Changelog
 
+## 0.21.1 (unreleased)
+- [[#129](https://github.com/IronCoreLabs/ironoxide/pull/129)] Improved error message for SDK initialization failure
```
giarc3: I think that even with a single entry, we've tended to put the PR and the description on two separate lines

src/internal/document_api/mod.rs
clintfred: This is a test. In all real instances there will be an underlying cause.

giarc3: I missed that. Great

Issue #131, opened 2020-03-27
Allow TLS implementation to be chosen via feature flag


clintfred:
AC:
* [x] support dynamically linking the target's OS's native TLS implementation (default)
* [x] support statically linking openssl for OS's that support openssl (Linux)
* [x] support using a rustls-tls for a pure rust implementation

Issue #130, opened 2020-03-19
Add more android targets


clintfred:
Let's add the following android targets!! I think this covers the non-deprecated android architectures
* aarch64-linux-android
* armv7-linux-androideabi
* x86_64-linux-android
* i686-linux-android (for the emulator)

You will want to use the one of the non-default TLS impls added in #132

clintfred:
It would also be nice to build test with all three TLS implementations from #132 on CI. We don't need to do the cross product of features with architectures. We can do all the features on x86_64.

PR #132, opened 2020-03-27
[#131] Add feature flags to configure TLS impl used

clintfred:
see #131 

This is not currently exercised in CI. It should be added with #130

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,6 @@ regex = "~1.3"
 ring = { version= "~0.16", features = ["std"] }
 recrypt = "~0.11"
 url= "~2.1.0"
-#reqwest = {version="~0.10.0", features = ["json"]}
 reqwest = {version="~0.10.0", features = ["json"], default-features = false}
 hex = "~0.3"
 itertools = "~0.9.0"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -60,6 +60,19 @@ all-features = true
...
+# enable to use statically compiled openssl on supported OpenSSL platforms; use with 'default-features = false'
+tls-vendored-openssl = ["reqwest/native-tls-vendored"]
+
+# enable to use rustls-tls; use with 'default-features = false'
+tls-rustls = ["reqwest/rustls-tls"]
+
+# dynamically use the target platform's native TLS
+tls-default = ["reqwest/default-tls"]
+
+default = ["tls-default"]
```
giarc3: Is there a reason this wouldn't be `default = ["reqwest/default-tls"]`?

coltfred:
Besides the tests we should have, I'm ok with this. Maybe add words to the github actions migration ticket so we add those tests when we move?

Cargo.toml
clintfred: No. They are equivalent. I wanted it to be clear that you need to choose one of the three `tls-` features, and once I defined it, I figured I might as well use it.

Issue #136, opened 2020-04-15
SEGV in cross test


cjyar:
```
Fiddle:ironoxide chris$ cross test --no-default-features --features "blocking beta tls-vendored-openssl" --target armv7-linux-androideabi
...
test crypto::aes::tests::test_parallel_encrypt ... libc: Fatal signal 11 (SIGSEGV), code 1, fault addr 0x813ceed4 in tid 82 (qemu-arm)
libc: Unable to open connection to debuggerd: Connection refused
qemu: uncaught target signal 11 (Segmentation fault) - core dumped
error: test failed, to rerun pass '--lib'

Caused by:
  process didn't exit successfully: `qemu-arm /target/armv7-linux-androideabi/debug/deps/ironoxide-c38a6b0d31c7f14c` (signal: 11, SIGSEGV: invalid memory reference)
```

This does not happen every time; it failed 8/10 times on my machine.

cjyar:
I got this on `aarch64-linux-android`, which seems even more concerning:
```
test crypto::aes::tests::test_parallel_encrypt ... libc: Fatal signal 11 (SIGSEGV), code 0, fault addr 0x0 in tid 21 ()
libc: Fatal signal 4 (SIGILL), code 2, fault addr 0x4004227030 in tid 23 (qemu-aarch64)
libc: Fatal signal 11 (SIGSEGV), code 1, fault addr 0x4802102498 in tid 22 (qemu-aarch64)
libclibc: Unable to open connection to debuggerd: Connection refused
: Unable to open connection to debuggerd: Connection refused
libc: Unable to open connection to debuggerd: Connection refused
qemu: uncaught target signal 4 (Illegal instruction) - core dumped
error: test failed, to rerun pass '--lib'

Caused by:
  process didn't exit successfully: `qemu-aarch64 /target/aarch64-linux-android/debug/deps/ironoxide-c26b43cb4deb804d` (signal: 4, SIGILL: illegal instruction)
```

clintfred:
Reading the `cross` docs:
> Also, testing is very slow. cross test runs units tests sequentially because QEMU gets upset when you spawn multiple threads. This means that, if one of your unit tests spawns threads, then it's more likely to fail or, worst, never terminate.

So that's probably the issue as the test in question is using `std::thread::spawn` which is maybe a bad idea anyway....

PR #137, opened 2020-04-15
Travis: Build several android architectures.

cjyar:
This PR doesn't build or run tests, because
- #136 
- Test builds don't complete on i686 or x86_64; it just hangs trying to build.

### DIFF
--- a/.travis_scripts/cross-test.sh
+++ b/.travis_scripts/cross-test.sh
@@ -5,7 +5,7 @@ set -x
 
 # If no special target provided, use default cargo arch for building and run all unit tests
 if [ -z "$TARGET" ]; then
-    IRONCORE_ENV=dev cargo t --features "blocking beta" --verbose
+    IRONCORE_ENV=dev cargo test --features "blocking beta" --verbose
     cargo fmt -- --check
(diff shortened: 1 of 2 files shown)


.travis_scripts/cross-test.sh
```diff
@@ -5,13 +5,13 @@ set -x
 
 # If no special target provided, use default cargo arch for building and run all unit tests
 if [ -z "$TARGET" ]; then
-    IRONCORE_ENV=dev cargo t --features "blocking beta" --verbose
+    IRONCORE_ENV=dev cargo test --features "blocking beta" --verbose
```
ernieturner, line 8: Was there a reason to change this?

cjyar: Craig and Clint said `cargo t` used to have special meaning, but it doesn't anymore. It seemed like a good idea to use the long form of the command in CI. If that's not true, I'll change it back.

ernieturner: Ya, that's sort of what I remembered as well which is why I called it out. I agree that if it doesn't have special meaning changing it to use the fully qualified command is better.

PR #139, opened 2020-04-16
Update DocumentOps docs

giarc3:
Trying to make ironoxide more easily understood from the docs alone. This PR focuses on the `document` functions and things related to them. 

I added examples for every `DocumentOps` function, which are compiled (but not run, because of async issues) when running `cargo test`, helping prove that they don't become out-of-date.

I need feedback on how the landing page of the docs site should flow. My current thought is that each section (Users, Groups, Documents), as well as a higher level (like Initialization) will provide a bare-bones example of the most important functions to get you up and running. A reader could start at the top and see how to initialize, create a user, create a group, and encrypt/decrypt a document right from that page. There will also be links to important functions for each section.

### DIFF
--- a/src/document/mod.rs
+++ b/src/document/mod.rs
@@ -298,9 +298,9 @@ pub trait DocumentOps {
     /// `grant_list` - List of users and groups to grant access to.
     ///
     /// # Errors
-    /// Fails only if the calling user does not have access to the document. If the response is an `Ok`, then the resulting
-    /// `DocumentAccessResult` will indicate which grants succeeded and which failed, and it will provide explanations for
-    /// each failure.
+    /// This operation supports partial success. If the request succeeds, then the resulting
+    /// `DocumentAccessResult` will indicate which grants succeeded and which failed, and it
+    /// will provide an explanation for each failure.
     ///
     /// # Examples
     /// ```
@@ -332,9 +332,9 @@ pub trait DocumentOps {
     /// `revoke_list` - List of users and groups to revoke access from.
     ///
     /// # Errors
-    /// Fails only if the calling user does not have access to the document. If the response is an `Ok`, then the resulting
-    /// `DocumentAccessResult` will indicate which revocations succeeded and which failed, and it will provide explanations for
-    /// each failure.
+    /// This operation supports partial success. If the request succeeds, then the resulting
+    /// `DocumentAccessResult` will indicate which revocations succeeded and which failed, and it
+    /// will provide an explanation for each failure.
     ///
     /// # Examples
     /// ```
--- a/src/document/mod.rs
+++ b/src/document/mod.rs
@@ -19,7 +19,7 @@ use itertools::{Either, EitherOrBoth, Itertools};
 /// Advanced document operations
 pub mod advanced;
 
-/// Explicit list of users and groups that should have access to decrypt a document.
+/// List of users and groups that should have access to decrypt a document.
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
 pub struct ExplicitGrant {
     grant_to_author: bool,
@@ -203,13 +203,13 @@ pub trait DocumentOps {
     /// # use ironoxide::prelude::*;
     /// # let sdk: IronOxide = unimplemented!();
     /// # let document_id: DocumentId = unimplemented!();
-    /// let document_data = sdk.document_get_metadata(&document_id).await?;
+    /// let document_meta = sdk.document_get_metadata(&document_id).await?;
     /// # Ok(())
     /// # }
     /// ```
     async fn document_get_metadata(&self, id: &DocumentId) -> Result<DocumentMetadataResult>;
 
-    /// Returns the document ID from the bytes of an IronCore encrypted document.
+    /// Returns the document ID from the bytes of an encrypted document.
     ///
     /// This is the same ID returned by `DocumentEncryptResult.id()`.
     ///
@@ -217,7 +217,7 @@ pub trait DocumentOps {
     /// - `encrypted_document` - Bytes of the encrypted document
     ///
     /// # Errors
-    /// Fails if the provided encrypted bytes have no header.
+    /// Fails if the provided bytes are not an encrypted document or have no header.
     ///
     /// # Examples
     /// ```
--- a/src/document/mod.rs
+++ b/src/document/mod.rs
@@ -42,8 +42,8 @@ impl ExplicitGrant {
 
 /// Parameters that can be provided when encrypting a new document.
 ///
-/// If no document ID is provided, one will be generated for it. If no document name is provided, the document's
-/// name will be left empty.
+/// Document IDs must be unique to the segment. If no ID is provided, one will be generated for it.
+/// If no document name is provided, the document's name will be left empty.
 ///
 /// For default parameters, use `DocumentEncryptOpts::default()`.
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
@@ -64,7 +64,7 @@ impl DocumentEncryptOpts {
     /// # Arguments
     /// - `id` - Unique ID to use for the document. Note: this ID will **not** be encrypted.
     /// - `name` - Non-unique name to use for the document. Note: this name will **not** be encrypted.
-    /// - `grants` - Grants that control who will be granted access to this encrypted document.
+    /// - `grants` - Grants that control who will have access to read and decrypt this document.
     pub fn new(
         id: Option<DocumentId>,
         name: Option<DocumentName>,
@@ -79,7 +79,7 @@ impl DocumentEncryptOpts {
     /// - `id` - Unique ID to use for the document. Note: this ID will **not** be encrypted.
     /// - `name` - Non-unique name to use for the document. Note: this name will **not** be encrypted.
     /// - `grant_to_author` - True if the calling user should have access to decrypt the document
-    /// - `grants` - List of users and groups that should have access to decrypt the document
+    /// - `grants` - List of users and groups that should have access to read and decrypt this document
     pub fn with_explicit_grants(
         id: Option<DocumentId>,
         name: Option<DocumentName>,
@@ -101,7 +101,7 @@ impl DocumentEncryptOpts {
     /// # Arguments
     /// - `id` - Unique ID to use for the document. Note: this ID will **not** be encrypted.
     /// - `name` - Non-unique name to use for the document. Note: this name will **not** be encrypted.
-    /// - `grants` - Policy to determine what users and groups will have access to the document.
+    /// - `grants` - Policy to determine which users and groups will have access to read and decrypt this document.
     pub fn with_policy_grants(
         id: Option<DocumentId>,
         name: Option<DocumentName>,
@@ -118,7 +118,8 @@ impl DocumentEncryptOpts {
 impl Default for DocumentEncryptOpts {
     /// Constructs a `DocumentEncryptOpts` with common values.
     ///
-    /// The document will have a generated ID and no name. Access to the document will be granted only to its creator.
+    /// The document will have a generated ID and no name. Only the document's author will be able to
+    /// read and decrypt it.
     fn default() -> Self {
         DocumentEncryptOpts::with_explicit_grants(None, None, true, vec![])
     }
@@ -128,9 +129,13 @@ impl Default for DocumentEncryptOpts {
 pub trait DocumentOps {
     /// Encrypts the provided document bytes.
     ///
-    /// This will create an entry for the document in the IronCore service. Returns a `DocumentEncryptResult`
-    /// which contains document metadata as well as the `encrypted_data`, which is the only thing that must be passed
-    /// to [document_decrypt()](trait.DocumentOps.html#tymethod.document_decrypt) in order to decrypt the document.
+    /// Returns a `DocumentEncryptResult` which contains document metadata as well as the `encrypted_data`,
+    /// which is the only thing that must be passed to [document_decrypt()](trait.DocumentOps.html#tymethod.document_decrypt)
+    /// in order to decrypt the document.
+    ///
+    /// Metadata about the document will be stored by IronCore, but the encrypted bytes of the document will not. To encrypt
+    /// without any document information being stored by IronCore, consider using
+    /// [document_encrypt_unmanaged()](advanced/trait.DocumentAdvancedOps.html#tymethod.document_encrypt_unmanaged) instead.
     ///
     /// # Arguments
     /// - `document_data` - Bytes of the document to encrypt
@@ -177,8 +182,7 @@ pub trait DocumentOps {
     /// ```
     async fn document_decrypt(&self, encrypted_document: &[u8]) -> Result<DocumentDecryptResult>;
 
-    /// Lists all of the IronCore encrypted documents that the calling user has been
-    /// granted access to.
+    /// Lists all of the encrypted documents that the calling user can read or decrypt.
     ///
     /// # Examples
     /// ```
@@ -192,7 +196,9 @@ pub trait DocumentOps {
     /// # }
     async fn document_list(&self) -> Result<DocumentListResult>;
 
-    /// Returns the metadata for an IronCore encrypted document.
+    /// Returns the metadata for an encrypted document.
+    ///
+    /// This will not return the encrypted document bytes, as they are not stored by IronCore.
     ///
     /// # Arguments
     /// - `id` - Unique ID of the document to retrieve
@@ -257,7 +263,7 @@ pub trait DocumentOps {
         new_document_data: &[u8],
     ) -> Result<DocumentEncryptResult>;
 
-    /// Updates a document's name to a new value.
+    /// Modifies or removes a document's name.
     ///
     /// Returns the updated metadata of the document.
     ///
@@ -283,13 +289,18 @@ pub trait DocumentOps {
         name: Option<&DocumentName>,
     ) -> Result<DocumentMetadataResult>;
 
-    /// Grants access to a document for a list of users and groups.
+    /// Grants decryption access to a document for the provided users and/or groups.
     ///
     /// Returns lists of successful and failed grants.
     ///
     /// # Arguments
     /// `document_id` - Unique ID of the document whose access is being modified.
     /// `grant_list` - List of users and groups to grant access to.
+    ///
+    /// # Errors
+    /// Fails only if the calling user does not have access to the document. If the response is an `Ok`, then the resulting
+    /// `DocumentAccessResult` will indicate which grants succeeded and which failed, and it will provide explanations for
+    /// each failure.
     ///
     /// # Examples
     /// ```
@@ -312,13 +323,18 @@ pub trait DocumentOps {
         grant_list: &Vec<UserOrGroup>,
     ) -> Result<DocumentAccessResult>;
 
-    /// Revokes access to a document for a list of users and groups.
+    /// Revokes decryption access to a document for the provided users and/or groups.
     ///
     /// Returns lists of successful and failed revocations.
     ///
     /// # Arguments
     /// `document_id` - Unique ID of the document whose access is being modified.
     /// `revoke_list` - List of users and groups to revoke access from.
+    ///
+    /// # Errors
+    /// Fails only if the calling user does not have access to the document. If the response is an `Ok`, then the resulting
+    /// `DocumentAccessResult` will indicate which revocations succeeded and which failed, and it will provide explanations for
+    /// each failure.
     ///
     /// # Examples
     /// ```
--- a/src/document/mod.rs
+++ b/src/document/mod.rs
@@ -19,16 +19,7 @@ use itertools::{Either, EitherOrBoth, Itertools};
 /// Advanced document operations
 pub mod advanced;
 
-/// Optional parameters that can be provided when encrypting a new document.
-#[derive(Clone, Debug, Eq, Hash, PartialEq)]
-pub struct DocumentEncryptOpts {
-    id: Option<DocumentId>,
-    name: Option<DocumentName>,
-    // at least one user/group must be included either explicitly or via a policy
-    grants: EitherOrBoth<ExplicitGrant, PolicyGrant>,
-}
-
-/// Explicit users/groups that should have access to decrypt a document.
+/// Explicit list of users and groups that should have access to decrypt a document.
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
 pub struct ExplicitGrant {
     grant_to_author: bool,
@@ -36,8 +27,11 @@ pub struct ExplicitGrant {
 }
 
 impl ExplicitGrant {
-    /// `grant_to_author` - true if the calling user should have access to decrypt the document
-    /// `grants` - other UserOrGroups that should have access to the document
+    /// Construct a new ExplicitGrant.
+    ///
+    /// # Arguments
+    /// - `grant_to_author` - True if the calling user should have access to decrypt the document
+    /// - `grants` - List of users and groups that should have access to decrypt the document
     pub fn new(grant_to_author: bool, grants: &[UserOrGroup]) -> ExplicitGrant {
         ExplicitGrant {
             grant_to_author,
@@ -46,7 +40,31 @@ impl ExplicitGrant {
     }
 }
 
+/// Parameters that can be provided when encrypting a new document.
+///
+/// If no document ID is provided, one will be generated for it. If no document name is provided, the document's
+/// name will be left empty.
+///
+/// For default parameters, use `DocumentEncryptOpts::default()`.
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct DocumentEncryptOpts {
+    id: Option<DocumentId>,
+    name: Option<DocumentName>,
+    // at least one user/group must be included either explicitly or via a policy
+    grants: EitherOrBoth<ExplicitGrant, PolicyGrant>,
+}
+
 impl DocumentEncryptOpts {
+    /// Constructs a new `DocumentEncryptOpts`.
+    ///
+    /// This requires either an `ExplicitGrant`, a `PolicyGrant`, or both. If only using one type
+    /// of grant, consider using [with_explicit_grants()](./struct.DocumentEncryptOpts.html#method.with_explicit_grants)
+    /// or [with_policy_grants()](./struct.DocumentEncryptOpts.html#method.with_policy_grants) instead.
+    ///
+    /// # Arguments
+    /// - `id` - Unique ID to use for the document. Note: this ID will **not** be encrypted.
+    /// - `name` - Non-unique name to use for the document. Note: this name will **not** be encrypted.
+    /// - `grants` - Grants that control who will be granted access to this encrypted document.
     pub fn new(
         id: Option<DocumentId>,
         name: Option<DocumentName>,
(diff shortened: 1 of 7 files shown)


src/document/mod.rs
```diff
@@ -46,14 +40,46 @@ impl ExplicitGrant {
     }
 }
 
+/// Parameters that can be provided when encrypting a new document.
+///
+/// If no document ID is provided, one will be generated for it. If no document name is provided, the document's
+/// name will be left empty.
```
ernieturner, line 46:
```suggestion
/// name will be left empty. Document IDs must be unique per segment.
```

src/document/mod.rs
```diff
@@ -46,14 +40,46 @@ impl ExplicitGrant {
...
+    /// Constructs a new `DocumentEncryptOpts`.
+    ///
+    /// This requires either an `ExplicitGrant`, a `PolicyGrant`, or both. If only using one type
+    /// of grant, consider using [with_explicit_grants()](./struct.DocumentEncryptOpts.html#method.with_explicit_grants)
+    /// or [with_policy_grants()](./struct.DocumentEncryptOpts.html#method.with_policy_grants) instead.
+    ///
+    /// # Arguments
+    /// - `id` - Unique ID to use for the document. Note: this ID will **not** be encrypted.
+    /// - `name` - Non-unique name to use for the document. Note: this name will **not** be encrypted.
+    /// - `grants` - Grants that control who will be granted access to this encrypted document.
```
ernieturner, line 67:
```suggestion
    /// - `grants` - Grants that control who will be granted access to decrypt this document.
```

src/document/mod.rs
```diff
@@ -84,113 +116,225 @@ impl DocumentEncryptOpts {
 }
 
 impl Default for DocumentEncryptOpts {
-    /// default to only sharing with the creator of the document
+    /// Constructs a `DocumentEncryptOpts` with common values.
+    ///
+    /// The document will have a generated ID and no name. Access to the document will be granted only to its creator.
```
ernieturner, line 121:
```suggestion
    /// The document will have a generated ID and no name. Access to decrypt the document will be granted only to its author.
```

src/document/mod.rs
```diff
@@ -84,113 +116,225 @@ impl DocumentEncryptOpts {
...
         DocumentEncryptOpts::with_explicit_grants(None, None, true, vec![])
     }
 }
 
 #[async_trait]
 pub trait DocumentOps {
-    /// List all of the documents that the current user is able to decrypt.
+    /// Encrypts the provided document bytes.
+    ///
+    /// This will create an entry for the document in the IronCore service. Returns a `DocumentEncryptResult`
```
ernieturner, line 131: > This will create an entry for the document in the IronCore service.

Reading this, I'm not entirely sure what it means or what we want to convey here. Maybe we should just remove it entirely? I can't think of a great way to reword it to make it more obvious what we mean. I mostly just don't want people to think that we're storing the encrypted document or anything.

src/document/mod.rs
```diff
@@ -84,113 +116,225 @@ impl DocumentEncryptOpts {
...
+    /// The document will have a generated ID and no name. Access to the document will be granted only to its creator.
     fn default() -> Self {
         DocumentEncryptOpts::with_explicit_grants(None, None, true, vec![])
     }
 }
 
 #[async_trait]
 pub trait DocumentOps {
-    /// List all of the documents that the current user is able to decrypt.
+    /// Encrypts the provided document bytes.
```
ernieturner, line 129:
```suggestion
    /// Encrypts the provided document bytes and optionally grants access to other users or groups to be able to decrypt the document.
```

src/document/mod.rs
```diff
@@ -84,113 +116,225 @@ impl DocumentEncryptOpts {
...
+    /// # Examples
+    /// ```
+    /// # async fn run() -> Result<(), ironoxide::IronOxideErr> {
+    /// # use ironoxide::prelude::*;
+    /// # let sdk: IronOxide = unimplemented!();
+    /// use ironoxide::document::DocumentEncryptOpts;
+    /// let data = "secret data".as_bytes();
+    /// let encrypted = sdk.document_encrypt(data, &DocumentEncryptOpts::default()).await?;
+    /// # Ok(())
+    /// # }
```
ernieturner, line 148: Just curious, what do the `#` prefixes mean vs the lines that don't have them?

src/document/mod.rs
```diff
@@ -84,113 +116,225 @@ impl DocumentEncryptOpts {
...
+    /// # let encrypted_data: Vec<u8> = vec![];
+    /// let decrypted_document = sdk.document_decrypt(&encrypted_data).await?;
+    /// let decrypted_data = decrypted_document.decrypted_data();
+    /// # Ok(())
+    /// # }
+    /// ```
+    async fn document_decrypt(&self, encrypted_document: &[u8]) -> Result<DocumentDecryptResult>;
+
+    /// Lists all of the IronCore encrypted documents that the calling user has been
+    /// granted access to.
```
ernieturner, line 181: Maybe we just remove the "IronCore" part here. I know that your goal here was that the unmanaged documents won't be returned here, but I'm not sure if that's worth specifically calling out? I dunno, I could be swayed either way, I was just trying to make it less vague as to what an "IronCore document" meant.
```suggestion
    /// Lists all of the encrypted documents that the calling user can decrypt
```

src/document/mod.rs
```diff
@@ -84,113 +116,225 @@ impl DocumentEncryptOpts {
...
+    /// # use ironoxide::document::DocumentListMeta;
+    /// # let sdk: IronOxide = unimplemented!();
+    /// let document_data = sdk.document_list().await?;
+    /// let documents: Vec<DocumentListMeta> = document_data.result().to_vec();
+    /// # Ok(())
+    /// # }
     async fn document_list(&self) -> Result<DocumentListResult>;
 
-    /// Get the metadata for a specific document given its ID.
+    /// Returns the metadata for an IronCore encrypted document.
```
ernieturner, line 195:
```suggestion
    /// Returns the metadata for an encrypted document. Does not return the encrypted document bytes.
```

src/document/mod.rs
```diff
@@ -84,113 +116,225 @@ impl DocumentEncryptOpts {
...
     ///
-    /// # Returns
-    /// `DocumentMetadataResult` with details about the requested document.
+    /// # Examples
+    /// ```
+    /// # async fn run() -> Result<(), ironoxide::IronOxideErr> {
+    /// # use ironoxide::prelude::*;
+    /// # let sdk: IronOxide = unimplemented!();
+    /// # let document_id: DocumentId = unimplemented!();
+    /// let document_data = sdk.document_get_metadata(&document_id).await?;
```
ernieturner, line 206:
```suggestion
    /// let document_meta = sdk.document_get_metadata(&document_id).await?;
```

src/document/mod.rs
```diff
@@ -84,113 +116,225 @@ impl DocumentEncryptOpts {
...
+    /// # Ok(())
+    /// # }
+    /// ```
     async fn document_get_metadata(&self, id: &DocumentId) -> Result<DocumentMetadataResult>;
 
-    /// Attempt to parse the document ID out of an encrypted document.
+    /// Returns the document ID from the bytes of an IronCore encrypted document.
```
ernieturner, line 212:
```suggestion
    /// Returns the document ID from the bytes of an encrypted document.
```

src/document/mod.rs
```diff
@@ -84,113 +116,225 @@ impl DocumentEncryptOpts {
...
+    ///
+    /// This is the same ID returned by `DocumentEncryptResult.id()`.
```
ernieturner, line 214:
```suggestion
    /// This is the same ID returned by `DocumentEncryptResult.id()` or `DocumentMetadataResult.id()`.
```

src/document/mod.rs
```diff
@@ -84,113 +116,225 @@ impl DocumentEncryptOpts {
...
     ///
     /// # Arguments
-    /// - `encrypted_document` - Encrypted document bytes
+    /// - `encrypted_document` - Bytes of the encrypted document
+    ///
+    /// # Errors
+    /// Fails if the provided encrypted bytes have no header.
```
ernieturner, line 220:
```suggestion
    /// Fails if the provided bytes are not an encrypted document or have no header.
```

src/document/mod.rs
```diff
@@ -84,113 +116,225 @@ impl DocumentEncryptOpts {
...
+    /// # let sdk: IronOxide = unimplemented!();
+    /// # let bytes: Vec<u8> = vec![];
+    /// let document_id = sdk.document_get_id_from_bytes(&bytes)?;
+    /// # Ok(())
+    /// # }
+    /// ```
     fn document_get_id_from_bytes(&self, encrypted_document: &[u8]) -> Result<DocumentId>;
 
-    /// Encrypt the provided document bytes.
+    /// Updates the contents of an existing IronCore encrypted document.
```
ernieturner, line 234:
```suggestion
    /// Updates the contents of an existing encrypted document while preserving the documents metadata and user and group access control.
```

src/document/mod.rs
```diff
@@ -84,113 +116,225 @@ impl DocumentEncryptOpts {
...
-    ///       `grants` - List of users/groups to grant access to this document once encrypted
-    async fn document_encrypt(
-        &self,
-        document_data: &[u8],
-        encrypt_opts: &DocumentEncryptOpts,
-    ) -> Result<DocumentEncryptResult>;
-
-    /// Update the encrypted content of an existing document. Persists any existing access to other users and groups.
+    /// The new contents will be encrypted, and which users and groups are granted access
+    /// will remain unchanged.
```
ernieturner, line 237: Removing as this has been combined with the above.
```suggestion
```

src/document/mod.rs
```diff
@@ -84,113 +116,225 @@ impl DocumentEncryptOpts {
...
         &self,
         id: &DocumentId,
         new_document_data: &[u8],
     ) -> Result<DocumentEncryptResult>;
 
-    /// Decrypts the provided encrypted document and returns details about the document as well as its decrypted bytes.
-    ///
-    /// # Arguments
-    /// - `encrypted_document` - Bytes of encrypted document. Should be the same bytes returned from `document_encrypt`.
+    /// Updates a document's name to a new value.
```
ernieturner, line 260:
```suggestion
    /// Modify or remove a document's name.
```

src/document/mod.rs
```diff
@@ -84,113 +116,225 @@ impl DocumentEncryptOpts {
...
+    /// # }
+    /// ```
     async fn document_update_name(
         &self,
         id: &DocumentId,
         name: Option<&DocumentName>,
     ) -> Result<DocumentMetadataResult>;
 
-    /// Grant access to a document. Recipients of document access can be either users or groups.
+    /// Grants access to a document for a list of users and groups.
```
ernieturner, line 286:
```suggestion
    /// Grants decryption access to a document for the provided users and/or groups.
```

src/document/mod.rs
```diff
@@ -84,113 +116,225 @@ impl DocumentEncryptOpts {
...
     /// # Arguments
-    /// `document_id` - id of the document whose access is is being modified
-    /// `grant_list` - list of grants. Elements represent either a user or a group.
+    /// `document_id` - Unique ID of the document whose access is being modified.
+    /// `grant_list` - List of users and groups to grant access to.
     ///
-    /// # Returns
-    /// Outer result indicates that the request failed either on the client or that the server rejected
-    /// the whole request. If the outer result is `Ok` then each individual grant to a user/group
-    /// either succeeded or failed.
```
ernieturner, line 177: I think this content is still pretty useful. It's worth telling users that this operation won't fail if all of the provided users/groups fail.

src/document/mod.rs
```diff
@@ -84,113 +116,225 @@ impl DocumentEncryptOpts {
...
+    /// # }
+    /// ```
     async fn document_grant_access(
         &self,
         document_id: &DocumentId,
         grant_list: &Vec<UserOrGroup>,
     ) -> Result<DocumentAccessResult>;
 
-    /// Revoke access from a document. Revocation of document access can be either users or groups.
+    /// Revokes access to a document for a list of users and groups.
```
ernieturner, line 315:
```suggestion
    /// Revokes decryption access to a document for the provided users and/or groups.
```

src/document/mod.rs
giarc3: The uniqueness is called out in the functions to create a DocumentEncryptOpts, so I would lean toward not including it here. Do you think it's needed in both places?
Or I guess I could also remove it from the functions and _only_ have it here?

src/document/mod.rs
giarc3: I worded it this way to try to imply that you're granted a little more than just the ability to decrypt (like seeing the ID, when it was last updated, who else has access, etc.). But you could probably argue that that extra information isn't worth implying. Thoughts?

src/document/mod.rs
giarc3: The `#` hides the line from the example. So boilerplate code that will be included in all of them (like the `async fn run`) or things where we're faking the typechecker on (like the `unimplemented!()`) won't have to show up for the users.

src/document/mod.rs
giarc3: Yeah I want a way to make that distinction, but I'm fine with taking it out for clarity.

This is one place where "that the calling user can decrypt" technically isn't correct though (which is why I wrote a test for `document_list` in the PR). If you're the author of a document, even if you make `grant_to_author` false, it will still show up in your list.

src/document/mod.rs
giarc3: While the addition is true, I was only calling out that it's the same as `DocumentEncryptResult.id()` to show that someone who just encrypted a file shouldn't call this, since they already have access to that data. Maybe I need to explicitly call that out.

So I think I'd rather leave out the addition to keep it focused on the intent of it.

src/document/mod.rs
giarc3: Rust documentation guidelines seem to want a brief summary of the function followed by additional notes/information on the following lines. So even though it would flow fine as a single sentence, I think it makes it too long for a brief summary.

src/document/mod.rs
giarc3: See above

src/document/mod.rs
giarc3: That's a good point. I think I'll make a `# Errors` section for this one to call that out

src/document/mod.rs
giarc3: I like it, though I'll change the verb tenses to match the standard/consistency.

src/document/mod.rs
ernieturner: I'm fine not including it here, but I'd update the line in the other places to say more than just "Unique ID". Having clarity of what we mean by "unique" would be good I think since it only has to be unique within the segment.

src/document/mod.rs
ernieturner: I suppose we could change it to say "who will have access to read and decrypt this document" or something like that, maybe?

src/document/mod.rs
ernieturner: Ah right, that weird corner case. I'd probably be ok with ignoring that here. Alternatively something like "Lists all of the encrypted documents that the caller user can decrypt or authored" I suppose. We can wait for someone else to weigh in too.

src/document/mod.rs
ernieturner: Yep, that's fine.

src/document/mod.rs
ernieturner: Ok, 👍

src/document/mod.rs
giarc3: I decided to have it in both places, since I agree that it makes it more clear

src/document/mod.rs
giarc3: :+1:

src/document/mod.rs
giarc3: Reworded slightly, but took the suggestion.

I agree that "author" makes the most sense for describing who made the document, but we should keep in mind that `AssociationType` refers to it as `Owner`, which conveys something slightly different

src/document/mod.rs
giarc3: The intent of this was to differentiate it from unmanaged encryption, but I agree that it doesn't convey that very well. I think it's important to call out that we're storing _some_ information about their document, but clarify that the data is not stored.

I'll try to work on the wording

src/document/mod.rs
giarc3: I think I'll try "Lists all of the encrypted documents that the calling user can read or decrypt."

src/document/mod.rs
giarc3: I think I'll go with
```
/// Returns the metadata for an encrypted document.
///
/// This will not return the encrypted document bytes, as they are not stored by IronCore.
```

src/document/mod.rs
giarc3: Done

src/document/mod.rs
giarc3: Done

src/document/mod.rs
giarc3: Done, but I could use some feedback on it

src/document/mod.rs
giarc3: Did it. Let me know what you think

src/document/mod.rs
```diff
@@ -84,113 +116,241 @@ impl DocumentEncryptOpts {
...
-    /// `grant_list` - list of grants. Elements represent either a user or a group.
+    /// `document_id` - Unique ID of the document whose access is being modified.
+    /// `grant_list` - List of users and groups to grant access to.
     ///
-    /// # Returns
-    /// Outer result indicates that the request failed either on the client or that the server rejected
-    /// the whole request. If the outer result is `Ok` then each individual grant to a user/group
-    /// either succeeded or failed.
+    /// # Errors
+    /// Fails only if the calling user does not have access to the document. If the response is an `Ok`, then the resulting
```
ernieturner, line 301: Given that the request to ironcore-id can also fail, this is a little simplified. It's maybe not worthwhile to try and enumerate all of the reasons it can fail. What if we said something like

```
This operation supports partial success. If any of the provided grants fail the `DocumentAccessResult` will indicate which grants succeeded and which failed, and it will provide explanations for each failure.
```

src/document/mod.rs
ernieturner: Now you made me sad that we aren't consistent with `author` and `owner` everywhere :(

src/document/mod.rs
giarc3: I added the part about partial success, removed the over-simplification, and tweaked the second sentence

src/document/mod.rs
```diff
@@ -84,113 +116,241 @@ impl DocumentEncryptOpts {
...
+    /// # let sdk: IronOxide = unimplemented!();
+    /// # let encrypted_data: Vec<u8> = vec![];
+    /// let decrypted_document = sdk.document_decrypt(&encrypted_data).await?;
+    /// let decrypted_data = decrypted_document.decrypted_data();
+    /// # Ok(())
+    /// # }
+    /// ```
+    async fn document_decrypt(&self, encrypted_document: &[u8]) -> Result<DocumentDecryptResult>;
+
+    /// Lists all of the encrypted documents that the calling user can read or decrypt.
```
clintfred, line 185:
```suggestion
    /// Lists metadata for all of the encrypted documents that the calling user can read/decrypt.
```

src/document/mod.rs
```diff
@@ -84,113 +116,241 @@ impl DocumentEncryptOpts {
...
     ///
-    /// # Returns
-    /// `DocumentMetadataResult` with details about the requested document.
+    /// # Examples
+    /// ```
+    /// # async fn run() -> Result<(), ironoxide::IronOxideErr> {
+    /// # use ironoxide::prelude::*;
+    /// # let sdk: IronOxide = unimplemented!();
+    /// # let document_id: DocumentId = unimplemented!();
+    /// let document_meta = sdk.document_get_metadata(&document_id).await?;
```
clintfred, line 212: Not sure how much value this example has. Doesn't show me anything I can't get from looking at the signature.

src/document/mod.rs
```diff
@@ -84,113 +116,241 @@ impl DocumentEncryptOpts {
...
 
-    /// Grant access to a document. Recipients of document access can be either users or groups.
+    /// Grants decryption access to a document for the provided users and/or groups.
+    ///
+    /// Returns lists of successful and failed grants.
     ///
     /// # Arguments
-    /// `document_id` - id of the document whose access is is being modified
-    /// `grant_list` - list of grants. Elements represent either a user or a group.
+    /// `document_id` - Unique ID of the document whose access is being modified.
```
clintfred, line 297: bullet these to get them to appear on separate lines in HTML

src/document/mod.rs
```diff
@@ -84,113 +116,241 @@ impl DocumentEncryptOpts {
...
 
-    /// Revoke access from a document. Revocation of document access can be either users or groups.
+    /// Revokes decryption access to a document for the provided users and/or groups.
+    ///
+    /// Returns lists of successful and failed revocations.
     ///
     /// # Arguments
-    /// `document_id` - id of the document whose access is is being modified
-    /// `revoke_list` - List of revokes. Elements represent either a user or a group.
+    /// `document_id` - Unique ID of the document whose access is being modified.
```
clintfred, line 331: bullet here

src/document/mod.rs
```diff
@@ -84,113 +116,241 @@ impl DocumentEncryptOpts {
...
+    /// ```
+    /// # async fn run() -> Result<(), ironoxide::IronOxideErr> {
+    /// # use ironoxide::prelude::*;
+    /// # let sdk: IronOxide = unimplemented!();
+    /// # let document_id: DocumentId = unimplemented!();
+    /// # let users: Vec<UserId> = vec![];
+    /// use std::convert::TryFrom;
+    /// use ironoxide::document::UserOrGroup;
+    /// // from a list of UserIds, `users`
+    /// let users_or_groups: Vec<UserOrGroup> = users.iter().map(|user| user.into()).collect();
```
clintfred, line 315: This is a great example as the [UserId] -> [UserOrGroup] conversion can be confusing. It also makes me want to make that easier, somehow...

src/document/mod.rs
```diff
@@ -84,113 +116,241 @@ impl DocumentEncryptOpts {
...
+    /// # }
+    /// ```
     async fn document_update_name(
         &self,
         id: &DocumentId,
         name: Option<&DocumentName>,
     ) -> Result<DocumentMetadataResult>;
 
-    /// Grant access to a document. Recipients of document access can be either users or groups.
+    /// Grants decryption access to a document for the provided users and/or groups.
```
clintfred, line 292: I wonder if we should talk more about partial success failure in the main comment block. Both here and for revoke.

src/document/mod.rs
```diff
@@ -200,28 +360,6 @@ pub trait DocumentOps {
 
 #[async_trait]
 impl DocumentOps for crate::IronOxide {
-    async fn document_list(&self) -> Result<DocumentListResult> {
```
clintfred, line 203: What are these changes? Reordering maybe?

src/document/mod.rs
giarc3: I agree that it can be confusing. Even worse is if you have a [UserId] and a [GroupId] and want to combine them into a [UserOrGroup]. I think it requires two mutable vecs and multiple lines. 
I would also love to make that process easier.

src/document/mod.rs
giarc3: Yes, re-ordering. The order we put in the file changes the order of the docs, so I tried to make a logical progress of functions that started with encrypt/decrypt

src/document/mod.rs
```diff
@@ -46,14 +40,46 @@ impl ExplicitGrant {
...
+    id: Option<DocumentId>,
+    name: Option<DocumentName>,
+    // at least one user/group must be included either explicitly or via a policy
+    grants: EitherOrBoth<ExplicitGrant, PolicyGrant>,
+}
+
 impl DocumentEncryptOpts {
+    /// Constructs a new `DocumentEncryptOpts`.
+    ///
+    /// This requires either an `ExplicitGrant`, a `PolicyGrant`, or both. If only using one type
```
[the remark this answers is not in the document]
giarc3: Done

src/document/mod.rs
```diff
@@ -46,14 +40,46 @@ impl ExplicitGrant {
...
+    /// of grant, consider using [with_explicit_grants()](./struct.DocumentEncryptOpts.html#method.with_explicit_grants)
```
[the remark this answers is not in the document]
giarc3: Done

src/document/mod.rs
```diff
@@ -46,14 +40,46 @@ impl ExplicitGrant {
...
+    /// or [with_policy_grants()](./struct.DocumentEncryptOpts.html#method.with_policy_grants) instead.
```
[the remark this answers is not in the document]
giarc3: Done

src/document/mod.rs
```diff
@@ -70,6 +96,12 @@ impl DocumentEncryptOpts {
         }
     }
 
+    /// Constructs a new `DocumentEncryptOpts` with access granted by a policy.
+    ///
+    /// # Arguments
+    /// - `id` - Unique ID to use for the document. Note: this ID will **not** be encrypted.
+    /// - `name` - Non-unique name to use for the document. Note: this name will **not** be encrypted.
+    /// - `grants` - Policy to determine which users and groups will have access to read and decrypt this document.
```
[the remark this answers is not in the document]
giarc3: Fixed the error and linked to the policy module

src/document/mod.rs
```diff
@@ -84,113 +116,241 @@ impl DocumentEncryptOpts {
...
 
 #[async_trait]
 pub trait DocumentOps {
-    /// List all of the documents that the current user is able to decrypt.
+    /// Encrypts the provided document bytes.
     ///
-    /// # Returns
-    /// `DocumentListResult` struct with vec of metadata about each document the user can decrypt.
+    /// Returns a `DocumentEncryptResult` which contains document metadata as well as the `encrypted_data`,
+    /// which is the only thing that must be passed to [document_decrypt()](trait.DocumentOps.html#tymethod.document_decrypt)
```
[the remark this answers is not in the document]
giarc3: I've gone back on forth on my preference. I went ahead and removed the `()` from our links in every file

src/document/mod.rs
```diff
@@ -84,113 +116,241 @@ impl DocumentEncryptOpts {
...
+    /// # Arguments
+    /// - `document_data` - Bytes of the document to encrypt
+    /// - `encrypt_opts` - Document encryption parameters. Default values are provided with `DocumentEncryptOpts::default()`.
+    ///
+    /// # Examples
+    /// ```
+    /// # async fn run() -> Result<(), ironoxide::IronOxideErr> {
+    /// # use ironoxide::prelude::*;
+    /// # let sdk: IronOxide = unimplemented!();
+    /// use ironoxide::document::DocumentEncryptOpts;
```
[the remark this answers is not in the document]
giarc3: Sure. Hidden

src/document/mod.rs
giarc3: I added the "metadata for", but I prefer to keep the "or" over the slash

src/document/mod.rs
giarc3: We discussed this elsewhere, but to recap, I'd prefer to leave this in as I think it gives an "alternate summary" of how to easily use the function
I also expanded this example slightly by showing the creation of the `DocumentId` from a string

src/document/mod.rs
giarc3: I think that the "Returns lists of successful and failed grants" implies the partial success, and the `Errors` section clarifies it

src/document/mod.rs
giarc3: Good catch. Fixed

src/document/mod.rs
giarc3: Fixed

PR #140, opened 2020-04-17
Expand prelude

giarc3:
Added to the prelude so it includes everything that's needed for initialization

If there are more things that would make sense to add, let me know

Cargo fmt also played with Colt's search imports, so sorry about that

### DIFF
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -1,12 +1,15 @@
+#[doc(no_inline)]
 pub use crate::{
+    config::{IronOxideConfig, PolicyCachingConfig},
     document::DocumentOps,
     group::GroupOps,
     internal::{
         document_api::{DocumentId, DocumentName},
         group_api::{GroupId, GroupName},
         user_api::{DeviceId, DeviceName, UserId},
+        DeviceContext, DeviceSigningKeyPair, IronOxideErr, PrivateKey,
     },
     policy::PolicyGrant,
     user::UserOps,
-    DeviceContext, IronOxide, IronOxideErr,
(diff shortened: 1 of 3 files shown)


src/prelude.rs
```diff
@@ -1,12 +1,15 @@
...
+    config::{IronOxideConfig, PolicyCachingConfig},
     document::DocumentOps,
     group::GroupOps,
     internal::{
         document_api::{DocumentId, DocumentName},
         group_api::{GroupId, GroupName},
         user_api::{DeviceId, DeviceName, UserId},
+        DeviceContext, DeviceSigningKeyPair, IronOxideErr, PrivateKey,
     },
     policy::PolicyGrant,
```
giarc3, line 12: I'm not sure why this is here, but removing it would make this a breaking change

src/prelude.rs
```diff
@@ -1,12 +1,15 @@
+#[doc(no_inline)]
```
giarc3, line 1: This makes the docs categorize these as re-exports.

Before, when searching the docs for "IronOxideErr", you'd get two results, `ironoxide::IronOxideErr` and `ironoxide::prelude::IronOxideErr`

Now it will only return `ironoxide::IronOxideErr` as intended

ernieturner: That's awesome, I've been spending some time in the docs while doing the Swift stuff and noticed that weirdness. Thanks for adding it.

giarc3:
Closing this PR because I'm working on one that will make bigger changes #141

PR #142, opened 2020-04-21
Rework exports/prelude

giarc3:
Trying to make our export better/easier to understand from the docs page.

Changes:
- Made the prelude a "kitchen sink" that re-exports all of our types. It's intended for someone who wants to wildcard import one thing and be done with it.
- Fixed several types that we're exported anywhere (DeviceId, DeviceName) or were exported only in the prelude (DocumentId, DocumentName).
- Added a `common` module that just changes where some types get exported. Things like `DeviceContext` and `PrivateKey` live here, making the landing page of the docs site more focused. I'd love to move more things from the top level to here (like `PrivateKeyRotationCheck`), but it's not really true that that's a "common" type.

Explanations for some individual changes are commented throughout the PR.

### DIFF
--- a/src/blocking/prelude.rs
+++ b/src/blocking/prelude.rs
@@ -1,2 +1,0 @@
-#[doc(no_inline)]
-pub use crate::{blocking::*, prelude::*};
--- a/src/blocking/prelude.rs
+++ b/src/blocking/prelude.rs
@@ -1,0 +1,2 @@
+#[doc(no_inline)]
+pub use crate::{blocking::*, prelude::*};
--- a/src/internal/mod.rs
+++ b/src/internal/mod.rs
@@ -5,7 +5,7 @@
 use crate::internal::{
     group_api::GroupId,
     rest::{Authorization, IronCoreRequest, SignatureUrlString},
-    user_api::{DeviceId, DeviceName, UserId},
+    user_api::UserId,
 };
 use chrono::{DateTime, Utc};
 use futures::Future;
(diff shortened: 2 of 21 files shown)


src/internal/mod.rs
```diff
@@ -417,60 +417,6 @@ impl From<DeviceAddResult> for DeviceContext {
     }
 }
 
-#[derive(Clone, Debug, Eq, Hash, PartialEq)]
-pub struct DeviceAddResult {
```
giarc3, line 421: This was moved to the `user_api` module as it's only ever used with user operations

src/internal/mod.rs
```diff
@@ -676,29 +622,6 @@ impl From<AugmentationFactor> for RecryptPrivateKey {
     }
 }
 
-/// Public/Private asymmetric keypair that is used for decryption/encryption.
-#[derive(Clone, Debug, Eq, Hash, PartialEq)]
-pub struct KeyPair {
```
giarc3, line 681: This was also moved to `user_api` because it's only used for user operations

src/internal/user_api/mod.rs
```diff
@@ -320,6 +343,53 @@ pub async fn user_rotate_private_key<CR: rand::CryptoRng + rand::RngCore>(
     .into())
 }
 
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct DeviceAddResult {
+    /// The user's given id, which uniquely identifies them inside the segment.
+    pub(crate) account_id: UserId,
```
giarc3, line 349: These were made `pub(crate)` so that the values can be accessed from the `impl From<DeviceAddResult> for DeviceContext`, which is in `ironoxide::internal`

src/lib.rs
```diff
@@ -109,6 +105,15 @@ use vec1::Vec1;
 pub type Result<T> = std::result::Result<T, IronOxideErr>;
 type PolicyCache = DashMap<PolicyGrant, Vec<WithKey<UserOrGroup>>>;
 
+// This is where we export structs that don't fit into a single module.
+// They were previously exported at the top level, but added clutter to the docs landing page.
+/// Types that are useful in multiple modules
+pub mod common {
+    pub use crate::internal::{
+        DeviceContext, DeviceSigningKeyPair, PrivateKey, PublicKey, SdkOperation,
```
giarc3, line 113: I would like to get more top-level types into here (like `PrivateKeyRotationCheckResult`), but that's not really a "common" struct, so we'd need to re-word/purpose the module.

src/prelude.rs
```diff
@@ -1,12 +1,5 @@
...
-    group::GroupOps,
-    internal::{
-        document_api::{DocumentId, DocumentName},
-        group_api::{GroupId, GroupName},
-        user_api::{DeviceId, DeviceName, UserId},
-    },
-    policy::PolicyGrant,
-    user::UserOps,
-    DeviceContext, IronOxide, IronOxideErr,
+    common::*, config::*, document::advanced::*, document::*, group::*, policy::*, user::*,
```
giarc3, line 3: I was debating between wildcard exports and individual exports, but this will be more maintainable was we add new things

src/prelude.rs
```diff
@@ -1,12 +1,5 @@
...
+    InitAndRotationCheck, IronOxide, IronOxideErr, PrivateKeyRotationCheckResult,
```
giarc3, line 4: I can't wildcard re-export the top-level exports because that would force our custom `Result` type on anyone who imported the prelude.

src/blocking/prelude.rs
```diff
@@ -0,0 +1,2 @@
+#[doc(no_inline)]
+pub use crate::{blocking::*, prelude::*};
```
giarc3, line 2: An alternative to this file would be to just have the top-level prelude export `blocking::*` with conditional compilation.

src/prelude.rs
```diff
@@ -1,12 +1,5 @@
+#[doc(no_inline)]
```
giarc3, line 1: The `search` module currently isn't included in this "kitchen sink" prelude. It could be with conditional compilation?

src/blocking/prelude.rs
giarc3: Went with that approach

src/prelude.rs
giarc3: Done

src/internal/user_api/mod.rs
clintfred: I think it would be somewhat cleaner just to move the `From` into the user_api module. We still need to think about potentially opening up all our *Result types, making all members pub.

clintfred:
PR description still mentions blocking prelude.

Do we want to include any examples of using the prelude, especially in the blocking.rs module?

giarc3:
> I don't see anything obviously wrong here, and I like the overall goals. This is a pretty big change to the public API, so we may want to try updating some downstream dependencies to see how the new structure feels before merging this.

In ironoxide-bindings, it was actually very simple. We're already importing the prelude in `lib.rs`, so I just simplified the imports to `use ironoxide::{prelude::*, blocking::BlockingIronOxide as IronOxide};` and all three workspaces compiled successfully

> PR description still mentions blocking prelude.
>
> Do we want to include any examples of using the prelude, especially in the blocking.rs module?

I removed the mention of the blocking prelude. What sort of example do you mean? I think it's in a pretty good place, as the decision of whether to import the prelude is the same whether the user is using blocking or async

clintfred:
Don't forget a changelog entry.

PR #148, opened 2020-04-30
Add serde for EncryptedBlindIndexSalt

BobWall23:


tests/search_ops.rs
```diff
@@ -19,6 +19,18 @@ mod search_tests {
...
     }
 
+    #[tokio::test]
+    async fn serde_blind_index_salt_roundtrips() -> Result<(), IronOxideErr> {
+        let ebis = EncryptedBlindIndexSalt {
+            encrypted_deks: vec![0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15],
+            encrypted_salt_bytes:vec![0, 2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 22, 24, 26, 28, 30]};
+        let ebis_str = serde_json::to_string(&ebis).unwrap();
+        let ebis2: EncryptedBlindIndexSalt = serde_json::from_str(&ebis_str).unwrap();
+        assert_eq!(ebis2.encrypted_deks, ebis.encrypted_deks);
```
clintfred, line 29: `EncryptedBlindIndexSalt` implements `PartialEq` so you can compare `ebis` and `ebis2` directly.

tests/search_ops.rs
```diff
@@ -19,6 +19,18 @@ mod search_tests {
...
     }
 
+    #[tokio::test]
+    async fn serde_blind_index_salt_roundtrips() -> Result<(), IronOxideErr> {
+        let ebis = EncryptedBlindIndexSalt {
+            encrypted_deks: vec![0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15],
+            encrypted_salt_bytes: vec![0, 2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 22, 24, 26, 28, 30],
+        };
+        let ebis_str = serde_json::to_string(&ebis).unwrap();
+        let ebis2: EncryptedBlindIndexSalt = serde_json::from_str(&ebis_str).unwrap();
```
coltfred, line 29: Do you want to assert about what this string looks like? Sometimes it's good to understand what the actual json would look like in case we want to do renames of items.

ernieturner: Without more serde annotations, won't the JSON keys be serialized as `encrypted_deks` and `encrypted_salt_bytes`? Should we camelCase those?

BobWall23: I would actually prefer producing an opaque blob of bytes, but I figured serde might give consumers more options. I thought about asserting on the JSON, but I don't know how stable the generated JSON is. I'd be bummed to have to fix tests because we updated to a new version of serde and it put an extra space after a comma or something.

BobWall23: I didn't really care about the case of the JSON keys, and I found guidelines that said to use underscores, but now I see other guidelines that say use camelCase - I guess that's the Google thing. Pushed a change to camelCase

PR #150, opened 2020-05-09
Expose transliterate_string method

BobWall23:


src/search.rs
```diff
@@ -169,6 +169,17 @@ impl BlindIndexSearch {
...
     }
+
+    /// Process a string using the same transliteration that is done when a string is processed to
+    /// tokenize data or a query. This processing will drop any characters that are ignored (mostly
+    /// punctuation), remove any diacritical marks, convert characters to sequences of latin equivalents,
+    /// and lower case all characters.
+    /// For example, the string “Æneid” is converted to “aeneid”, and “北亰” is converted to “bei jing ”.
+    ///
+    /// s - The string you want to process
+    pub fn transliterate_string(&self, s: &str) -> String {
```
coltfred, line 180: Seems like since this doesn't use BlindIndexSearch that we should instead just `pub use` the `transliterate_string` and allow it to be exposed out in the search module?

BobWall23: Yup - fixed and made sure the docs retrieved the info from search helpers for the function.

tests/INTEGRATION-TESTS.md
```diff
@@ -34,7 +34,7 @@ IronOxide tests can be run against any other environment, with proper setup. To
 
 ### Identity Assertion Key File
 
-An Identity Assertion Key file must be downloaded from the admin console interface immediately after creating a new Identity Assertion Key. It must be named `iak.pem` and placed in `./tests/testkeys/`.
+An Identity Assertion Key file must be downloaded from the admin console interface immediately after creating a new Identity Assertion Key. It must be named `iak-${IRONCORE_ENV}.pem` (defaults to `iak-prod.pem` if the IRONCORE_ENV variable is not set) and placed in `./tests/testkeys/`.
```
giarc3, line 37: This isn't accurate. This markdown section is for testing against a different environment (not dev, stage, or prod), when they choose to set `IRONCORE_ENV=htttps://some-url.com`. In this case, the file must be named exactly `iak.pem`. 

It's true that it defaults to prod when the variable is unset, but that shouldn't go into the `Testing against a different environment` section if we want to add that in. And since there are default configs/keys encrypted for IronCore use, I personally don't think that we need to call this out.

tests/INTEGRATION-TESTS.md
```diff
@@ -50,7 +50,7 @@ An IronCore Config file can be downloaded from the admin console on creation of
 
 Note that case is significant for the key names.
 
-This file must be named `ironcore-config.json` and placed in `./tests/testkeys/`.
+This file must be named `ironcore-config-${IRONCORE_ENV}.json` and placed in `./tests/testkeys/` (`ironcore-config-prod.json` if IRONCORE_ENV is not set).
```
giarc3, line 53: Same comment as above

tests/INTEGRATION-TESTS.md
BobWall23: I see - I jumped into the "other environments" section and thought the IAK always needed to be in `iak.pem`. I'll revert this.

PR #154, opened 2020-05-20
Improve docs for document types

giarc3:
Improve the documentation for all of the `document` module types. 

Also added some convenience `From/TryFrom` implementations we were missing (`DocumentName`, `UserOrGroup`)

### DIFF
--- a/src/document/advanced.rs
+++ b/src/document/advanced.rs
@@ -1,3 +1,7 @@
+//! Advanced document types
+//!
+//! See [DocumentAdvancedOps](trait.DocumentAdvancedOps.html) for advanced document functions and key terms.
+
 pub use crate::internal::document_api::{
     DocumentDecryptUnmanagedResult, DocumentEncryptUnmanagedResult,
 };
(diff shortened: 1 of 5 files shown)


clintfred:
Should there be some changes in the `lib.rs` "Document Operations" section, similar to what we did in the "User Operations"?

giarc3:
> Should there be some changes in the `lib.rs` "Document Operations" section, similar to what we did in the "User Operations"?

I haven't gotten to the "User Operations" section yet, but I did a re-write of "Document Operations" last ticket.

src/document/mod.rs
```diff
@@ -1,3 +1,7 @@
+//! Document types
```
clintfred, line 1: Be aware that this gets combined with the comment block in `lib.rs` on `pub mod document`. I think I slightly prefer to put the docs here. We could just make the ones in `lib.rs` `//` comments, or probably just remove them.

src/document/advanced.rs
```diff
@@ -1,3 +1,7 @@
+//! Advanced document types
+//!
+//! See [DocumentAdvancedOps](trait.DocumentAdvancedOps.html) for advanced document functions and key terms.
```
clintfred, line 3: I like these links to the trait at the top of the module. Helps people find the functions more easily.

src/document/advanced.rs
```diff
@@ -1,3 +1,7 @@
+//! Advanced document types
```
clintfred, line 1: I don't think I like the word "types" here. Maybe "Advanced document operations"? I think it would also be good to put some text here describing the tradeoff between managed and unmanaged encryption.

src/document/advanced.rs
```diff
@@ -9,37 +13,39 @@ use crate::{
 };
 use itertools::EitherOrBoth;
 
+/// IronOxide Advanced Document Operations
+///
+/// # Key Terms
+/// - ID     - The ID representing a document. It must be unique within the document's segment and will **not** be encrypted.
+/// - Name   - The human-readable name of a document. It does not need to be unique and will **not** be encrypted.
+/// - EDEKs  - Encrypted document encryption keys used by unmanaged document encryption and decryption.
```
clintfred, line 21: EDEKS are produced by unmanged document encryption and used by unmanaged document decryption

src/document/advanced.rs
```diff
@@ -9,37 +13,39 @@ use crate::{
 };
 use itertools::EitherOrBoth;
 
+/// IronOxide Advanced Document Operations
+///
+/// # Key Terms
+/// - ID     - The ID representing a document. It must be unique within the document's segment and will **not** be encrypted.
```
clintfred, line 19: Do we need to mention ID and Name here? They aren't used in this trait, right?

src/document/mod.rs
clintfred: image

src/document/mod.rs
```diff
@@ -40,17 +43,16 @@ impl ExplicitGrant {
 /// Parameters that can be provided when encrypting a new document.
 ///
 /// Document IDs must be unique to the segment. If no ID is provided, one will be generated for it.
-/// If no document name is provided, the document's name will be left empty.
+/// If no name is provided, the document's name will be left empty.
 ///
-/// For default parameters, use `DocumentEncryptOpts::default()`.
+/// Default values are provided with `DocumentEncryptOpts::default()`.
```
clintfred, line 48: I think it would be nice to link to the Default impl here as that is where we describe what the defaults are.

clintfred:
> > Should there be some changes in the `lib.rs` "Document Operations" section, similar to what we did in the "User Operations"?
> 
> I haven't gotten to the "User Operations" section yet, but I did a re-write of "Document Operations" last ticket. This PR is just filling in the things I forgot to do from that ticket (though you did notice those missing parts at the time).

Gotcha. I am just saying that as part of this second round of cleanup, it might be nice to think about putting some more links, at least to the DocumentOps trait, in the lib.rs comment block. If a user is looking at the main page of the crate, it might help them find where they want to go. Or maybe we want to rework the main doc page altogether? I'm just trying to think of ways to get people into the right section of the docs more quickly.

src/document/mod.rs
giarc3: Sorry, this is a result of my bad "git-ness". I had some changes in the group PR and some in the document PR, and did a bad job of separating them. I believe the group PR removes all of the `lib.rs` `///` comments and switches to individual file comments. 
I'll be sure it all merges together correctly.

src/document/advanced.rs
giarc3: I chose "types" because this isn't where you find the operations, that's `DocumentAdvancedOps`. This file is just 2 structs and a trait. So I'm good with changing the wording, but I'd rather leave "operations" for the trait itself.

src/document/advanced.rs
giarc3: Well `id` and `name` are fields of `DocumentEncryptOpts`, which is shared for both kinds of encryption even if it maybe shouldn't be. 
And `id` is returned as part of `DocumentEncryptUnmanagedResult` and `DocumentDecryptUnmanagedResult`.
So it seems like we should at least have some mention here.

src/document/advanced.rs
clintfred: I guess I kind of see your point from the prospective of the doc site since the *Result types get re-exported from the `internal` module. The main thing in the module is a trait which doesn't fit my definition of a "type". Maybe "Advanced Document functionality" or "Advanced Document API"?

src/document/advanced.rs
clintfred: Since they aren't directly applicable, I would remove them. Maybe some of the explanation about uniqueness and encryption should be on `DocumentEncryptOps`?

src/document/advanced.rs
giarc3: I went with "Advanced document API".
Similar change done for the regular document API.

src/document/advanced.rs
giarc3: There is a sentence about ID uniqueness there. I've added the part about not encrypting them there and removed those two bullet points from `DocumentAdvancedOps`.

src/document/advanced.rs
giarc3: Fixed

src/document/mod.rs
giarc3: Good point. Did that here and in the arguments for `document_encrypt`/`document_encrypt_unmanaged`

PR #159, opened 2020-05-29
Added example of encrypted search patterns

BobWall23:


examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
+// =============================================================================
```
clintfred, line 1: Do we have a stable link to the pattern that this sample demonstrates?

I think it would be good to try to give a very abbreviated description of what this example is illustrating. 
This sentence from the pattern seems like a good start.
> The following implementation patterns illustrate how you can use the IronCore SDKs in your application to do short substring searches of data that is end-to-end encrypted.

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
...
+use mut_static::MutStatic;
+use std::{
+    convert::TryFrom,
+    fmt,
+    fs::File,
+    path::PathBuf,
+    time::{SystemTime, UNIX_EPOCH},
+};
+
+type Result<T> = std::result::Result<T, AppErr>;
```
clintfred, line 30: You might consider looking at https://github.com/dtolnay/anyhow to reduce the error handling code in this sample. I think it would allow you to get rid of all the `From` instances at the bottom of the file.

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
...
+// Also note that this program creates two different groups - in order to avoid
+// errors when executing the program multiple times, it appends the current
+// time (as a time_t) when the program is started to the group IDs.
+//
+// Copyright (c) 2020  IronCore Labs, Inc.
+// =============================================================================
+
+use ironoxide::prelude::*;
+use lazy_static::lazy_static;
+use mut_static::MutStatic;
```
clintfred, line 21: Looks like this is being used for GLOBAL_STATE. We normally wouldn't recommend using this type of pattern in a real application, so I'm wondering if we can either do something else or perhaps explain why we're using it here (stand in for persistent storage?)

BobWall23: You didn't think these comments were sufficient?
```
// Mock out some functions that would call the back-end service and return data in the real app
// We use some mutable global state to persist things that the service usually would.
```

clintfred: I did see those later on. They do help.

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
...
+    // Create a group to use for the encryption of customer records.
+    // Note this is not the same as the group that was used to secure
+    // the blind index, but anyone that needs to be able to search for
+    // customers by name will need to belong to both groups.
+    let group_id = create_group_id("customerService")?;
+    create_group(&sdk, &group_id, "Customer Service").await?;
+
+    // Add some test customers to the "server store"
+    let cust1 = encrypt_customer(&sdk, 1, "Gumby", "", &group_id).await?;
+    save_customer(cust1, &[], &[]);
```
coltfred, line 74: It took me a while to figure out why the users which are saved with no index tokens would be useful for this example. Either make a separate function which has documentation about why that's useful or calculate the tokens.

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
...
+    let cust2 = encrypt_customer(&sdk, 2, "Æ neid 北亰", "", &group_id).await?;
+    save_customer(cust2, &[], &[]);
+    let cust3 = encrypt_customer(&sdk, 3, "aeneid bei jing", "", &group_id).await?;
+    save_customer(cust3, &[], &[]);
+
+    // Allow the user to enter additional customers
+    let mut next_id = 4;
```
coltfred, line 81: It seems like breaking this out into a separate function would make this more readable. The reading of the values isn't key to the example.

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
...
+            email = email.trim().to_string();
+
+            let cust = encrypt_customer(&sdk, next_id, &name, &email, &group_id).await?;
+            save_customer(cust, &[], &[]);
+        }
+        next_id += 1;
+    }
+
+    // now create a new customer record to index
+    let mut customer = Customer {
```
coltfred, line 106: It seems like the idea of having a single Customer type which represents both the encrypted version of the Customer and the decrypted version of the Customer wouldn't be advisable. How do you tell if the name is encrypted or not?

This would also allow you to remove the email_keys and name_keys from the Customer type and just have them in the EncryptedCustomer type.

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
...
+        None,                  // document ID - create unique
+        None,                  // document name
+        false,                 // don't encrypt to self
+        vec![group_id.into()], // users and groups to which to grant access
+    );
+    let enc_name = sdk
+        .document_encrypt_unmanaged(customer.name.as_bytes(), &encrypt_opts)
+        .await?;
+    // Replace name with encoded encrypted version. Also need to store EDEKs to decrypt name.
+    customer.name = base64::encode(enc_name.encrypted_data());
```
coltfred, line 131: You wouldn't need to base64 encode this if you used a type that had `bytes` for the encrypted name.

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
...
+    customer.name_keys = base64::encode(enc_name.encrypted_deks());
+    save_customer(customer.clone(), &name_tokens, &[]);
```
coltfred, line 133: Why didn't you compute the index for the email and save them as well? You did it for customer 2, but not this one.

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
...
+        id: 24,
+        name: "Pokey".to_string(),
+        name_keys: "".to_string(),
+        email: "pokey@gumby.io".to_string(),
+        email_keys: "".to_string(),
+    };
+
+    // start-snippet{useSecondIndex}
+    // Generate the index tokens for the customer name and email address, then encrypt them
+    let name_tokens = blind_index
```
coltfred, line 159: This is the same thing you just did above, it'd be better if we refactored it out to take a `Customer` and returned a `EncryptedCustomer` with the 2 sets of index tokens.

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
...
+            .to_string(),
+        ))
+    }
+}
+
+// Add the program start time to the end of the specified string to
+// form a unique id string, then create a GroupId from it
+// start-snippet{createGroupId}
+fn create_group_id(id_str: &str) -> Result<GroupId> {
+    let stime = &GLOBAL_STATE.read().unwrap().start_time;
```
coltfred, line 206: I don't think time should be in GLOBAL_STATE, this is the only place it's used. Just pass it in and store it as state in the main.

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
...
+    let encrypt_opts = DocumentEncryptOpts::with_explicit_grants(
+        None,                  // document ID
+        None,                  // document name
+        true,                  // encrypt to self
+        vec![group_id.into()], // users and groups to which to grant access
+    );
+    let enc_result = sdk
+        .document_encrypt_unmanaged(name.as_bytes(), &encrypt_opts)
+        .await?;
+    let enc_name = base64::encode(enc_result.encrypted_data());
```
coltfred, line 253: I don't think there is much advantage to having these as strings. You can store them in the fake database regardless and this just confuses the types.

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
...
+    let enc_name_keys = base64::encode(enc_result.encrypted_deks());
+
+    Ok(Customer {
+        id,
+        name: enc_name,
+        name_keys: enc_name_keys,
+        email: email.to_string(),
+        email_keys: "".to_string(),
```
coltfred, line 261: Maybe this is the same question as before, but why aren't you encrypting the email?

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
...
+// words, decrypt the name in the customer record, then check whether the
+// decrypted name contains all of the word fragments from the query string.
+// If so, return the decrypted name. Otherwise, the record was a false positive,
+// so return None.
+// start-snippet{filterCust}
+async fn filter_customer(
+    sdk: &IronOxide,
+    cust: &Customer,
+    name_parts: &Vec<&str>,
+) -> Result<Option<String>> {
```
coltfred, line 275: Based on the function name, I would assume this should return the whole customer. I'd also suggest adding name to the function name. Maybe `ensure_customer_name_matches` or something.

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
...
+// Given a customer record and the transliterated query string, broken into
+// words, decrypt the name in the customer record, then check whether the
+// decrypted name contains all of the word fragments from the query string.
+// If so, return the decrypted name. Otherwise, the record was a false positive,
+// so return None.
+// start-snippet{filterCust}
+async fn filter_customer(
+    sdk: &IronOxide,
+    cust: &Customer,
+    name_parts: &Vec<&str>,
```
coltfred, line 274: This function should be able to take the `name_query` and then transliterate it inside here. Then the snippet in the docs can comment about why were splitting on whitespace and comparing. It wasn't immediately obvious to me so some comments would be useful.

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
...
+        }
+    }
+    Ok(())
+}
+// end-snippet{displayCust}
+
+// Mock out some functions that would call the back-end service and return data in the real app
+// We use some mutable global state to persist things that the service usually would.
+struct GlobalState {
+    pub encrypted_salt: String,
```
coltfred, line 326: You could probably save yourself the json serialize/deserialize and just put the EncryptedBlindIndex in there.

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
...
+    pub customers: Vec<Customer>,
+    pub start_time: String,
```
coltfred, line 328: This is only ever read and only in 1 function. I think it shouldn't be here.

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
...
+}
+
+impl GlobalState {
+    pub fn new() -> Self {
```
coltfred, line 332: If you want an empty version of a struct the more rust way would be to define Default for that struct.

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
...
+                .duration_since(UNIX_EPOCH)
+                .unwrap()
+                .as_secs()
+                .to_string(),
+        }
+    }
+}
+
+lazy_static! {
+    static ref GLOBAL_STATE: MutStatic<GlobalState> = MutStatic::new();
```
coltfred, line 346: If it were up to me, I would just pass this state to the functions where it's needed. The MutStatic adds more noise than it's worth IMO.

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,421 @@
...
+    GLOBAL_STATE.read().unwrap().encrypted_salt.clone()
+}
+
+// For now, we ignore the tokens associated with each customer rec
+fn save_customer(customer: Customer, _name_tokens: &[u32], _email_tokens: &[u32]) {
+    GLOBAL_STATE.write().unwrap().customers.push(customer);
+}
+
+// Just return all the records that we have in the store for now
+fn search_customers(_tokens: &Vec<u32>) -> Vec<Customer> {
```
coltfred, line 363: If we're trying to make this look similar to a real application it seems like we shouldn't just return all the values. This causes all the records they've added to always be returned giving the impression that there's no way to filter before you decrypt. 

To fix this all we have to do is change the state to be a `Map<Vec<u32>, Customer>>` and filter based on all the tokens matching.

examples/enc-search-sample.rs
BobWall23: Maybe in a future PR. I borrowed this from `ironoxide-cli`.

examples/enc-search-sample.rs
BobWall23: Used HashSets for the tokens.

examples/enc-search-sample.rs
BobWall23: Figured out a better initialization.

examples/enc-search-sample.rs
BobWall23: Done

examples/enc-search-sample.rs
BobWall23: Moved it out.

examples/enc-search-sample.rs
BobWall23: done

examples/enc-search-sample.rs
BobWall23: Restructured the save so it always computes the index

examples/enc-search-sample.rs
BobWall23: Used `Vec<u8}>`

examples/enc-search-sample.rs
BobWall23: Done

examples/enc-search-sample.rs
BobWall23: Done

examples/enc-search-sample.rs
BobWall23: Calculated the tokens - restructured save methods to do it in there instead of before calling.

examples/enc-search-sample.rs
BobWall23: Done

examples/enc-search-sample.rs
BobWall23: Done

examples/enc-search-sample.rs
```diff
@@ -0,0 +1,427 @@
...
+        enc_email: enc_email.encrypted_data().to_vec(),
+        email_keys: enc_email.encrypted_deks().to_vec()
+    };
+    save_customer_to_app_server(enc_cust, name_tokens, vec![]);
+    // end-snippet{indexData}
+
+    Ok(())
+}
+
+async fn save_customer2(
```
coltfred, line 217: It seems like we should comment why there are 2 save_customers and what is different about them. Alternatively, just give them function a more descriptive name.

coltfred (commented):
I would rename save_customer2, but other than that it looks much better.

coltfred (approved):

examples/enc-search-sample.rs
BobWall23: Done.

PR #161, opened 2020-06-05
Switch to GitHub actions

giarc3:
### Github Action Workflow
- Runs `cross build` for the same architectures as we did on Travis, and `cargo build` for `aarch64-apple-ios` because it can't use `cross`. These use features `beta, blocking, tls-rustls`.
- Runs the tests on Mac and Ubuntu, with 1 retry if the tests fail. These use features `beta, blocking`.
- Runs `cargo fmt` and fails the workflow if that fails.
- Runs the security audit daily and on change to `Cargo.toml`.

### To Reviewers:
Since this is a public repo and we don't have to pay for minutes, should we expand our matrix to run the tests with more combinations of features? We currently aren't testing: using just default features, just one or the other of `beta, blocking`, or the feature `tls-vendored-openssl`.


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,94 @@
...
+      matrix:
+        os: [ubuntu-18.04, macos-10.15]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+      - name: Generate Cargo.lock so we can cache from it
+        run: cargo generate-lockfile
+      - uses: actions/cache@v2
```
cjyar, line 23: Do we need the workaround from https://github.com/IronCoreLabs/recrypt-rs/blob/master/.github/workflows/ci.yaml#L17? I believe it was failing to actually restore the cache without that workaround, in certain environments. The cache action would print a zillion warning messages in the Actions log, but it would still succeed.
If we don't need that here, then maybe we can remove it from the other repos where we use it.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,94 @@
...
+          restore-keys: |
+            ${{ github.workflow }}-rust-${{ matrix.os }}-
+      - name: Decrypt test keys
+        run: openssl aes-256-cbc -d -K ${{ secrets.OPENSSL_KEY }} -iv ${{ secrets.OPENSSL_IV }} -in tests/testkeys/iak-dev.pem.enc -out tests/testkeys/iak-dev.pem
+      - name: Cargo test (with one retry on failure)
+        uses: nick-invision/retry@v1
+        with:
+          timeout_minutes: 30
+          max_attempts: 2
+          command: IRONCORE_ENV=dev cargo test --features "blocking beta"
```
cjyar, line 39: Why do we need to retry? This is a code smell.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,94 @@
...
+    continue-on-error: true
+    strategy:
+      matrix:
+        target:
+          [
+            aarch64-linux-android,
+            armv7-linux-androideabi,
+            i686-linux-android,
+            x86_64-linux-android,
+          ]
```
cjyar, line 52: Probably clearer to use YAML list syntax here, instead of JSON list syntax.
JSON list: `[a,b,c]`
YAML list:
```
- a
- b
- c
```

.github/workflows/ci.yaml
giarc3: It should no longer be needed. They released a fix for the issue a few weeks ago.
We should be able to remove it from our other repos (and switch them all to cache@v2), but I assume it's not worth making all the PRs without other changes.

.github/workflows/ci.yaml
giarc3: My thinking here was there we have a fairly large build matrix and a somewhat decent chance that one of the tests will fail due to timeout, internet issues, or similar (and this did occur during my testing). When this failure occurs, it will fail the whole workflow, and there's currently no way to restart individual jobs in a workflow (though many people have asked for that feature, so I have to believe it will come eventually).
So I added a single retry for the test step.
I'm okay with removing it, but if we do I anticipate we'll have some frustration here in the future.
Thoughts?

.github/workflows/ci.yaml
giarc3: Sure, I didn't know that was an option

.github/workflows/ci.yaml
cjyar: Very cool.

.github/workflows/ci.yaml
cjyar: I guess the real code smell here is that our "unit" tests have an external dependency that we're not mocking. Unless somebody wants to somehow mock ironcore-id, or stand up a local one for the tests, your workaround seems like the best option.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,96 @@
...
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ github.workflow }}-rust-${{ matrix.target }}-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ github.workflow }}-rust-${{ matrix.target }}-
+      - uses: actions-rs/cargo@v1
+        with:
+          # `cross` won't work on aarch64-apple-ios, so it falls back to `cargo` in that case
```
clintfred, line 78: we say we aren't using `cross`, but then the next line says `use-cross: true`? What am I missing?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,96 @@
...
+jobs:
+  cargo-test:
+    runs-on: ${{ matrix.os }}
+    continue-on-error: true
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, macos-10.15]
+        include:
+          - os: ubuntu-18.04
+            features: --features "blocking beta"
```
clintfred, line 18: what are these features vs the ones spelled out in the  uses: actions-rs/cargo@v1 block below?

giarc3: The block you mention is in a different job, so unrelated. But I'll be updating our matrix to cover features better in my next commit.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,96 @@
...
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ github.workflow }}-rust-${{ matrix.os }}-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ github.workflow }}-rust-${{ matrix.os }}-
+      - name: Decrypt test keys
+        run: openssl aes-256-cbc -d -K ${{ secrets.OPENSSL_KEY }} -iv ${{ secrets.OPENSSL_IV }} -in tests/testkeys/iak-dev.pem.enc -out tests/testkeys/iak-dev.pem
```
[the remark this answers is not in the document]
giarc3: We currently can only use our ironhide action on Linux, so it seemed unnecessary to write 2 methods of decrypting the test keys here. Once our ironhide action is more versatile, I agree that this should switch over.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,96 @@
...
+          command: IRONCORE_ENV=dev cargo test ${{ matrix.features }}
+
+  cargo-build:
+    runs-on: ${{ matrix.os }}
+    continue-on-error: true
+    strategy:
+      matrix:
+        target:
+          - aarch64-linux-android
+          - armv7-linux-androideabi
```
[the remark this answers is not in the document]
giarc3: Works for me. Replacing this with `x86_64-unknown-linux-musl`.

If we're good with removing this here, I assume we could also remove that target from our aar?

.github/workflows/ci.yaml
giarc3: We are using `cross`, it's just that in the `aarch64-apple-ios` case, using cross fails so it uses `cargo` instead.
I'll try to make this comment more clear.

.github/workflows/ci.yaml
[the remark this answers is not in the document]
giarc3: After more talking with you, I added back the armv7, but I'll keep in the musl as well.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,115 @@
...
+          - os: ubuntu-18.04
+            features: --features tls-rustls --no-default-features
+          - os: ubuntu-18.04
+            features: --features tls-default --no-default-features
+          # MSRV
+          - os: ubuntu-18.04
+            toolchain: 1.39.0
+    steps:
+      - uses: actions/checkout@v2
+      - if: matrix.toolchain != ''
```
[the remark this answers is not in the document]
giarc3: These machines actually already have Rust on them. So the only time I _really_ need to call the `actions-rs/toolchain` action is when I need something different than what they have (like if I need `cargofmt` or the `beta` toolchain). So this conditional makes it so that I don't have to add `toolchain: stable` to every different "include" build as it already has stable installed. 

As a side note, the reason that line 19 has `toolchain: [stable]` instead of skipping it is so that lines 34 and 35 _add_ to the build matrix instead of just modifying the ubuntu build. The `-include` is overloaded to do both, so it can lead to some confusing builds.

Issue #163, opened 2020-06-09
Replace dev depencency `frank_jwt` with `jsonwebtoken`


clintfred:
As a developer running the tests, 
I want an up to date, maintained, JWT library with minimal dependencies.

https://github.com/GildedHonour/frank_jwt is not being very actively maintained and it has a direct dependency on openssl. The openssl dependency makes running on windows require us to statically link openssl to compile the tests.

https://github.com/Keats/jsonwebtoken does not depend on openssl (using ring, instead, which we already depend on).

Directly related to #162 
Could be related to #160

Issue #162, opened 2020-06-09
Windows can't run integration tests with tls-rustls or tls-default features


clintfred:
This was discovered on CI.

To reproduce (on windows): 
* `cargo test --features tls-rustls --no-default-features`
* `cargo test` (default features)

Both of these should work as rustls is pure rust and the default should use the windows TLS implementation.

Workaround for now is to use tls-vendored-openssl
`cargo test --features tls-vendored-openssl --no-default-features`

Once we have validated that all TLS combinations work, we should update CI to use the windows native (default) TLS as that is what what most Windows applications would probably want.

PR #164, opened 2020-06-11
Add public Jwt type

giarc3:
Switched from `frank_jwt` to `jsonwebtoken`.

There already was an internal `Jwt` type, so re-purposed it and made it public. It validates that the jwt has the correct payload fields and an allowed algorithm. `user_create`, `user_verify`, and `generate_new_device` all use the new `Jwt` type instead of `&str`.

Also fixes:
- #162 
- #163

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-18.04, macos-10.15, windows-2019]
-        features: [--features tls-vendored-openssl --no-default-features]
+        features: [--features ""]
         toolchain: [stable]
         include:
           # different features combinations
@@ -29,7 +29,7 @@ jobs:
           - os: ubuntu-18.04
             features: --features tls-rustls --no-default-features
           - os: ubuntu-18.04
-            features: --features tls-default --no-default-features
+            features: --features tls-vendored-openssl --no-default-features
           # MSRV
           - os: ubuntu-18.04
             toolchain: 1.39.0
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,6 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # different operating systems
         os: [ubuntu-18.04, macos-10.15, windows-2019]
         features: [--features ""]
         toolchain: [stable]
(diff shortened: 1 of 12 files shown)


src/internal/user_api.rs
```diff
@@ -257,9 +258,78 @@ impl UserDevice {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+struct Claims {
+    sub: String,
+    pid: usize,
+    sid: String,
+    kid: Option<usize>,
```
ernieturner, line 266: While this is optional on the ironcore-id side, I'd vote that we don't let people omit this as it's in their best interest to not to it. The `kid` being optional has probably never really been shown as an option in IronOxide as we changed it over to be required before this repo existed.

.github/workflows/ci.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-18.04, macos-10.15, windows-2019]
-        features: [--features tls-vendored-openssl --no-default-features]
+        features: [--features ""]
```
clintfred, line 18: This is just switching the default back to tls-default?

src/internal/user_api.rs
clintfred: I agree. I don't think there's any value in letting someone omit `kid`. It would be nice to eventually omit the code that allows ahat on the server side at some point.

src/internal/user_api.rs
```diff
@@ -257,9 +258,78 @@ impl UserDevice {
...
+
+    pub fn jwt(&self) -> &str {
+        self.0.as_str()
+    }
+
+    fn to_utf8(&self) -> Vec<u8> {
+        self.0.as_bytes().to_vec()
+    }
+}
+impl std::fmt::Display for Jwt {
```
clintfred, line 312: It would be nice for either `Debug` or `Display` to show the parsed JWT. This is probably the right impl for `Display` and `Debug` should be changed.

.github/workflows/ci.yaml
giarc3: Correct, and now our `include` builds test `tls-rustls`and `tls-vendored-openssl`.

src/internal/user_api.rs
giarc3: Done.
`Display` is just the jwt string slice, while `Debug` has the string slice, headers, and claims.

src/internal/user_api.rs
giarc3: Sounds good to me. Done

PR #169, opened 2020-06-19
Added an example for encrypting to a group and to a user

coltfred:


examples/encrypting.rs
```diff
@@ -0,0 +1,66 @@
...
+/// The group that's encrypted to is created each run, but the user is just the current user.
+#[tokio::main]
+async fn main() -> Result<()> {
+    let (device_context, sdk) =
+        initialize_sdk_from_file(&PathBuf::from("examples/example-ironoxide-user.json")).await?;
+    encrypt_to_group(&sdk).await?;
+    encrypt_to_user(&sdk, device_context.account_id()).await?;
+    Ok(())
+}
+///Create a random group and encrypt the super secret message to it.
```
giarc3, line 15: In the most nit-picky way possible, a space after the `///` would make this more consistent.

examples/encrypting.rs
```diff
@@ -0,0 +1,66 @@
...
+        let device_context_file = File::open(&device_path)?;
+        let device_context: DeviceContext = serde_json::from_reader(device_context_file)?;
+        let ironoxide = ironoxide::initialize(&device_context, &Default::default()).await?;
+        Ok((device_context, ironoxide))
+    } else {
+        panic!(format!(
+            "Couldn't open file {} containing DeviceContext",
+            device_path.display()
+        )
+        .to_string(),)
```
giarc3, line 59: This `to_string(),` could be removed.

examples/example-ironoxide-user.json
```diff
@@ -0,0 +1,6 @@
+{
+  "accountId": "example-ironoxide-user",
+  "segmentId": 1825,
+  "signingPrivateKey": "IW67bY+Yee6oLV1T/AlS82bSP7d9ZqDH3EhmRGJj9js06T9Rx2SjAI3FRUf9O7sflOJ61gDpXVrAO/sNfGZoXw==",
+  "devicePrivateKey": "W7DvCJvGsnXunza+qstf/FTLW42h/VpWN8Be/CkClhY="
+}
```
giarc3, line 6: Where is this device from?

examples/encrypting.rs
coltfred: The format! itself can actually be removed, thanks for making me think about that harder.

examples/example-ironoxide-user.json
coltfred: Generated using the production pem file in the tests. We can't comment the json file and keep it json. Do you think it's important to comment on that?

giarc3: No, I don't think we need to comment it. We've just recently been using a specific project for public devices in tests, and we haven't made ironoxide's public yet, so I was wondering where you decided to get it from.

examples/encrypting.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+/// Create a group, asking the SDK to generate a unique ID for it.
+async fn encrypt_to_group(sdk: &IronOxide) -> Result<DocumentId> {
+    // start-snippet{encryptToGroup}
+    let group_id = create_group(sdk).await?;
+    let message = "This is my secret which a whole group should see.";
+    let encrypted_result = sdk
+        .document_encrypt(
+            message.as_bytes(),
+            &DocumentEncryptOpts::with_explicit_grants(None, None, true, vec![(&group_id).into()]),
```
clintfred, line 23: An alternative would be to `group_id.clone()` here

coltfred: I would still need the into(), so I'll stick with the borrow.

examples/encrypting.rs
```diff
@@ -0,0 +1,66 @@
+use anyhow::Result;
+use ironoxide::prelude::*;
+use std::{fs::File, path::PathBuf};
+
+/// This is a very basic example of encrypting to a user_id and a group_id.
+/// The group that's encrypted to is created each run, but the user is just the current user.
+#[tokio::main]
+async fn main() -> Result<()> {
+    let (device_context, sdk) =
+        initialize_sdk_from_file(&PathBuf::from("examples/example-ironoxide-user.json")).await?;
```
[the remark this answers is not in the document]
coltfred: true.

PR #170, opened 2020-06-30
Handle namespace on JWT private claims

BobWall23:
Also changed type of a couple of claims from `usize` to `u32`


src/internal/user_api.rs
```diff
@@ -259,23 +259,34 @@ impl UserDevice {
...
+    pub pid: u32,
+    /// Segment ID
+    #[serde(alias = "http://ironcore/sid")]
+    pub sid: String,
+    /// Service key ID
+    #[serde(alias = "http://ironcore/kid")]
+    pub kid: u32,
+    /// User ID
+    #[serde(alias = "http://ironcore/uid")]
+    pub uid: Option<String>,
```
giarc3, line 289: I haven't seen this claim before. Is it only used for Auth0? Can we get an explanation in the comments? From what we have now, I can't tell the difference between this and `sub`.

BobWall23: Yes, I guess. Auth0 fills in `sub` with some GitHub ID, and it puts the user's email into the `uid` claim. `ironcore-id` currently overwrites `sub` with `uid` if it is present.

giarc3: This should go into the doc comments either on `uid` or the whole `JwtClaims` (or both)

PR #173, opened 2020-07-13
Add code sample of encrypting to policy, to be added to website. Fix sample docs on encrypting to a policy

ernieturner:


src/policy.rs
```diff
@@ -70,7 +70,7 @@
 //! ```
 //! Example:
 //! If the current user of the sdk is "alice@ironcorelabs" and the following PolicyGrant is evaluated,
-//! `PolicyGrant::new("PII".try_from()?, "INTERNAL".try_from()?, None, None)` will match the second-to-last rule
+//! `PolicyGrant::new(Some(Category::try_from("PII")?), Some(Sensitivity::try_from("INTERNAL")?), None, None)` will match the second-to-last rule
```
giarc3, line 73: Do you think this would be better as a doc example? Then it would be compiled whenever we run the tests.
I'm fine either way, but if you wanted to add it, it would be something like this: 
```
//! ```
//! # use ironoxide::prelude::*;
//! # fn run() -> Result<(), IronOxideErr> {
//! # use std::convert::TryInto;
//! PolicyGrant::new(Some("PII".try_into()?), Some("INTERNAL".try_into()?), None, None);
//! # Ok(())
//! # }
//! ```
```

src/policy.rs
```diff
@@ -70,7 +70,7 @@
...
 //! in the example policy, above and will return users: [joe@ironcorelabs, alice@ironcorelabs] and
```
giarc3, line 74: While you're here, could you get rid of that comma? It makes no sense.

src/policy.rs
ernieturner: Done, hopefully it turns out correctly in the docs.

src/policy.rs
ernieturner: Done

PR #177, opened 2020-08-27
Increase type_length_limit due to a change in Rust 1.46.0 

clintfred:
Rust 1.46.0 seems to be generating longer type names, or perhaps a change in one of our dependencies is causing this - I'm not sure. 
A limit of 1670000 was suggested by the compiler (and seemed to be enough), so I just rounded up to 2000000 to give us some headroom in case things internal names get longer again.

Also bumped a couple of dependencies.

The crate builds without the `type_length_limit` directive on Rust `1.45.2`, but does not on `1.46.0` or `1.47.0-nightly (2d8a3b918 2020-08-26)`

### DIFF
--- a/src/internal/user_api.rs
+++ b/src/internal/user_api.rs
@@ -312,7 +312,7 @@ impl Jwt {
         // This `dangerous_unsafe_decode` is safe because the server will do the actual
         // signature verification and validation. We just want to do a little initial validation
         // to catch issues earlier.
-        let token_data = jsonwebtoken::dangerous_unsafe_decode::<JwtClaims>(jwt)
+        let token_data = jsonwebtoken::dangerous_insecure_decode::<JwtClaims>(jwt)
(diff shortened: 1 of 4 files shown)


src/internal/user_api.rs
```diff
@@ -312,7 +312,7 @@ impl Jwt {
         // This `dangerous_unsafe_decode` is safe because the server will do the actual
         // signature verification and validation. We just want to do a little initial validation
         // to catch issues earlier.
-        let token_data = jsonwebtoken::dangerous_unsafe_decode::<JwtClaims>(jwt)
+        let token_data = jsonwebtoken::dangerous_insecure_decode::<JwtClaims>(jwt)
```
giarc3, line 315: Oh hey, I'm the one who renamed that function in `jsonwebtoken` and deprecated the old name. Guess I never remembered to use the new name here.

clintfred: I thought that was you :)

PR #183, opened 2020-09-29
Upgrade to rust-protobuf 2.17

clintfred:
This is a breaking change for any downstream consumers that also use rust-protobuf.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,7 @@ anyhow = "~1.0.31"
 
 [build-dependencies]
 protobuf-codegen-pure = "~2.17"
-itertools = "0.9.0"
+itertools = "~0.9.0"
 
 [package.metadata.docs.rs]
 all-features = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ lazy_static = "~1.4"
 chrono = { version = "~0.4", features = ["serde"] }
 percent-encoding="~2.1"
 log = "~0.4"
-protobuf = { version = "~2.14", features = ["with-bytes"] }
+protobuf = { version = "~2.17", features = ["with-bytes"] }
 vec1 = "~1.6.0"
 # ironoxide requires rt-threaded at runtime, but not at compile time
 tokio = { version = "~0.2.0", features = ["time", "rt-threaded"] }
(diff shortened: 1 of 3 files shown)


BobWall23 (approved):

build.rs
```diff
@@ -32,7 +33,16 @@ fn main() {
...
         .unwrap();
-    let new_contents = format!("mod proto {{pub mod {} {{ \n{}\n}}}}", name, contents);
+
+    // Work around for https://github.com/rust-lang/rust/issues/5472
+    // Introduced in rust-protobuf https://github.com/stepancheg/rust-protobuf/pull/495
+    // More discussion: https://github.com/stepancheg/rust-protobuf/pull/523#issuecomment-701026992
+    let filtered: String = contents
+        .lines()
+        .filter(|line| line.trim() != "#![rustfmt::skip]")
+        .join("\n");
```
giarc3, line 43: This maybe feels a little over-engineered with the introduction of itertools as a build dependency (though I admit I haven't fully thought through whether it's negative to include it since we already have it as a regular dependency).
Would it be any worse to just do `let filtered: String = contents.replace("#![rustfmt::skip]", "");`

coltfred: It's certainly not the same behavior, though I can't tell you if the source could legally contain the code that would actually make it break. 

I like the itertools solution much better since it's more clear what will happen to each line exactly. 

I hope that rust will only build itertools once since it's matching the same version.

Cargo.toml
```diff
@@ -53,7 +53,8 @@ mut_static = "~5.0"
 anyhow = "~1.0.31"
 
 [build-dependencies]
-protobuf-codegen-pure = "~2.14"
+protobuf-codegen-pure = "~2.17"
+itertools = "0.9.0"
```
coltfred, line 57: Should this be ~0.9.0?

coltfred (changes requested):

build.rs
clintfred: I'm not too concerned about bringing in itertools at build time since I think we use it in nearly every rust codebase. It's like `std+` to me. Also, we either need to get off of rust-protobuf (prost) or figure out a better way of doing things longterm. In the short-term we need to get to 2.17 so we can use the "well-known-types" that google puts out for protobuf3.

Cargo.toml
clintfred: That would probably be better, yes. I'll change it.

build.rs
clintfred: I realize I didn't really address your concern in my last comment, @giarc3. Sorry. 

I think processing line by line is safer. There are certainly other ways of doing it, but I don't think it's worth fretting over it too much for this workaround.

giarc3 (approved):

coltfred (approved):

PR #193, opened 2020-10-26
Relax the semver requirement for protobuf

clintfred:
This will allow downstream consumers to choose their version of protobuf without causing build issues.

I realized this while looking at https://github.com/tikv/rust-prometheus/blob/master/Cargo.toml

I tested this by `git reset`ing logdriver to an old revision that used protobuf 2.14. It compiled file. It also compiled fine with a dependency of 2.17. 

There might still be problems in upgrading to newer version because we have custom `build.rs` logic for the proto generation, but at least cargo will be able to successfully select a version.

If we merge this, I think we can close https://github.com/IronCoreLabs/ironoxide/pull/188

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ lazy_static = "~1.4"
 chrono = { version = "~0.4", features = ["serde"] }
 percent-encoding="~2.1"
 log = "~0.4"
-protobuf = { version = "~2.17", features = ["with-bytes"] }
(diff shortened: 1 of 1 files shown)


giarc3 (approved):
My opinion is that `^` beats `~` for any dependency that we trust to conform to semver. 
Your change is equivalent to `^2`, which might be a little more clear? But maybe not.

coltfred:
Seems entirely reasonable to me.

PR #196, opened 2020-11-09
Allow benchmarks to be run on different environments

giarc3:
No longer hardcodes benchmarks to the `dev` environment. Adds 2 new `DeviceContext` files and selects on based on the `IRONCORE_ENV` that's set when running `cargo bench`.

Also adds a few extra benches made by Clint.

Also sorts `Cargo.toml` and adds a CI check to keep it sorted


BobWall23 (approved):
A bunch of check failures "functions in traits cannot be declared `async`" just appeared. Otherwise, looks good.

giarc3:
Not worrying about the failing tests as they seem intermittent and possibly related to our caching

clintfred (approved):
LGTM

PR #213, opened 2021-01-06
Upgrade tokio, reqwest, bytes

giarc3:
Also makes rust-protobuf use a minimum of v2.20 so that it's compatible with bytes v1.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ edition = "2018"
 async-trait = "0.1.21"
 base64 = "~0.13.0"
 base64-serde = "~0.6.1"
-bytes = "~0.5"
+bytes = "~0.6"
 chrono = { version = "~0.4", features = [ "serde" ] }
 dashmap = "4.0.1"
 futures = "~0.3.1"
@@ -34,12 +34,12 @@ rand = "~0.7"
 rand_chacha = "~0.2.2"
 recrypt = "~0.11"
 regex = "~1.4"
-reqwest = { version = "~0.10.0", features = [ "json" ], default-features = false }
+reqwest = { version = "~0.11.0", features = [ "json" ], default-features = false }
 ring = { version = "~0.16", features = [ "std" ] }
 serde = { version = "~1.0", features = [ "derive" ] }
 serde_json = "~1.0"
 # ironoxide requires rt-threaded at runtime, but not at compile time
-tokio = { version = "~0.2.0", features = [ "time", "rt-threaded" ] }
+tokio = { version = "~1.0", features = [ "time", "rt-multi-thread" ] }
 url = "~2.2.0"
 vec1 = "~1.6.0"
 
@@ -49,7 +49,7 @@ criterion = "~0.3"
 double = "~0.2.4"
 galvanic-assert = "~0.8"
 mut_static = "~5.0"
-tokio = { version = "~0.2.0", features = [ "macros" ] }
+tokio = { version = "~1.0", features = [ "macros" ] }
 uuid = { version = "~0.8", features = [ "v4" ], default-features = false }
 
 [build-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ edition = "2018"
 async-trait = "0.1.21"
 base64 = "~0.13.0"
 base64-serde = "~0.6.1"
-bytes = "~1"
+bytes = "~1.0"
 chrono = { version = "~0.4", features = [ "serde" ] }
 dashmap = "4.0.1"
 futures = "~0.3.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ criterion = "~0.3"
 double = "~0.2.4"
 galvanic-assert = "~0.8"
 mut_static = "~5.0"
-tokio = { version = "~1.0", features = [ "macros" ] }
+tokio = { version = "~1.0", features = [ "macros", "rt-multi-thread" ] }
 uuid = { version = "~0.8", features = [ "v4" ], default-features = false }
 
 [build-dependencies]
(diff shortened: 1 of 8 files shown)


coltfred (approved):

benches/ironoxide_bench.rs
```diff
@@ -119,80 +118,70 @@ fn criterion_benchmark(c: &mut Criterion) {
         group_enc_deks,
         group1,
         group2,
-    ) = rt.enter(|| block_on(f));
+    ) = rt.block_on(f);
 
     c.bench_function("document get metadata", |b| {
-        b.iter(|| rt.enter(|| block_on(io.document_get_metadata(black_box(enc_result.id())))))
+        b.iter(|| rt.block_on(io.document_get_metadata(black_box(enc_result.id()))))
```
giarc3, line 124: I don't remember why we used the `rt.enter` syntax in the first place, so if this change to `rt.block_on` is worse/wrong, please let me know

Cargo.toml
```diff
@@ -26,20 +26,19 @@ jsonwebtoken = "~7.2"
 lazy_static = "~1.4"
 log = "~0.4"
 percent-encoding = "~2.1"
-protobuf = { version = "^2.0", features = [ "with-bytes" ] }
+protobuf = { version = "^2.20", features = [ "with-bytes" ] }
```
clintfred, line 29: This is needed because this project isn't following semver and 2.19 isn't compatible?

Cargo.toml
```diff
@@ -26,20 +26,19 @@ jsonwebtoken = "~7.2"
...
 recrypt = "~0.11"
 regex = "~1.4"
-reqwest = { version = "~0.10.0", features = [ "json" ], default-features = false }
+reqwest = { version = "~0.11.0", features = [ "json" ], default-features = false }
 ring = { version = "~0.16", features = [ "std" ] }
 serde = { version = "~1.0", features = [ "derive" ] }
 serde_json = "~1.0"
-# ironoxide requires rt-threaded at runtime, but not at compile time
-tokio = { version = "~0.2.0", features = [ "time", "rt-threaded" ] }
+tokio = { version = "~1.0", features = [ "time", "rt-multi-thread" ] }
```
clintfred, line 41: Maybe we only need `rt-multi-threaded` for the `blocking` feature?

src/blocking.rs
```diff
@@ -251,47 +239,44 @@ impl BlockingIronOxide {
...
 
 /// Creates a tokio runtime with the default number of core threads (num of cores on a machine)
 /// and an elevated number of blocking_threads as we expect heavy concurrency to be network-bound
 fn create_runtime() -> tokio::runtime::Runtime {
-    tokio::runtime::Builder::new()
-        .threaded_scheduler() // use multi-threaded scheduler
+    tokio::runtime::Builder::new_multi_thread()
         .enable_all() // enable both I/O and time drivers
-        .max_threads(250) // core_threads default to number of cores, blocking threads are max - core
+        .max_blocking_threads(250) // worker_threads default to number of cores
```
clintfred, line 279: I'll need to do some reading on tokio 1.0's threading model, but in the past blocking threads wouldn't be what we wanted here...

benches/ironoxide_bench.rs
clintfred: I'm not sure either. I assume it didn't work the other way anymore?

clintfred (commented):
Left a few thoughts. I think my biggest concert in the blocking_threads config. I need to read more on that.

src/blocking.rs
coltfred: Well, this is off by the number of core threads, but I don't think that's worth arguing over. As the previous comment said setting max threads was setting blocking threads to 250-(core threads)

benches/ironoxide_bench.rs
giarc3: They at least changed it enough that it wouldn't compile as-is, but I didn't try too hard to make it work the same since `block_on` seemed more like what I expected anyway

src/blocking.rs
giarc3: I figured that we chose this number fairly arbitrarily in the first place, so it changing by a few while also being a bit more clear wouldn't be bad

Cargo.toml
giarc3: Correct. @coltfred had a discussion with the project owner, who didn't seem to care that the bytes upgrades were not compatible. So that's a bit worrisome going forward, but he did say that he hopes bytes will be compatible from now on.

coltfred: https://github.com/stepancheg/rust-protobuf/issues/539

Cargo.toml
giarc3: I think this is true? I was a little wary because of the comment that you had written about requiring `rt-threaded` at runtime. But if I depend on local ironoxide, ironoxide-cli still works fine (after a quick upgrade of tokio), so I think this means that it's all good to make it an optional flag?

giarc3: It looks like I can also remove the `time` flag altogether? Or is there some other test I'd have to do to check that

clintfred: If the tests and benches work without it, we don't need it.

skeet70: `rt-threaded` previously made all of tokio use the multi-threaded scheduler vs the single threaded one. Seems like `rt-multi-thread` is the new name of it, pretty sure we want it, but I don't know.

clintfred: We do need it for timeouts, so we added it back in. We were perhaps picking it up (transitively) via reqwest.

clintfred: As for `rt-multi-threaded`, there's nothing in the library that requires it, but it seems like most consumers would want it. I suggest we make it a default feature so it could be disabled if needed.

src/blocking.rs
clintfred: We aren't doing any spawn_blocking, so this can be removed.

clintfred:
This needs a changelog entry and then I think it's ready to merge.

clintfred (approved):

PR #215, opened 2021-01-08
Upgrade to recrypt 0.12.0

giarc3:


.github/workflows/ci.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
             features: --features tls-vendored-openssl --no-default-features
           # MSRV
           - os: ubuntu-18.04
-            toolchain: 1.39.0
+            toolchain: 1.41.1
```
clintfred, line 36: Matches recrypt's MSRV.

clintfred (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
   - Update to reqwest 0.11.0
   - Update to bytes 1.0
   - Require minimum protobuf of 2.20.0
+- [[#215](https://github.com/IronCoreLabs/ironoxide/pull/215)]
+  - Update to recrypt 0.12.0
+  - Bump MSRV to 1.41.1
```
BobWall23, line 12: Isn't this only pertinent to our build process? I think it might be best to omit that from the changelog.

BobWall23 (changes requested):

CHANGELOG.md
clintfred: MSRV applies to anyone depending on ironoxide. In other words, if you want to depend on ironoxide, you must be using rustc 1.41.1 or higher.

giarc3: Are you referring to the MSRV? That's the minimum version of rust that we say we support (and test against in CI), so I think it's important to include that for others to know.

BobWall23: OK - agreed.

BobWall23 (approved):

PR #216, opened 2021-01-14
Fix issue with serde 1.0.119

giarc3:
Breaking change in patch bump :upside_down_face: 

For some reason, we were implementing `Display` using an export of `serde`, but apparently we weren't supposed to be able to import that (they renamed the module to `__private`). This change makes us implement `Display` in the standard way using `std::fmt::Display`, so the change is backwards compatible and we don't have to set a minimum version of `serde`.


coltfred (approved):
Thanks for clarifying the reasoning in the PR. :+1:

PR #222, opened 2021-02-22
Change all dependencies to default upgrade requirements

giarc3:
Changes from `~` requirements to no symbol (which is equivalent to `^` requirements). From the docs:
> Caret requirements allow SemVer compatible updates to a specified version. An update is allowed if the new version number  > does not modify the left-most non-zero digit in the major, minor, patch grouping. In this case, if we ran cargo update -p time,  cargo should update us to version 0.1.13 if it is the latest 0.1.z release, but would not update us to 0.2.0. If instead we had specified the version string as ^1.0, cargo should update to 1.1 if it is the latest 1.y release, but not 2.0. The version 0.0.x is not considered compatible with any other version.
This compatibility convention is different from SemVer in the way it treats versions before 1.0.0. While SemVer says there is no compatibility before 1.0.0, Cargo considers 0.x.y to be compatible with 0.x.z, where y ≥ z and x > 0.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ reqwest = { version = "0.11", features = [ "json" ], default-features = false }
 ring = { version = "0.16", features = [ "std" ] }
 serde = { version = "1.0.123", features = [ "derive" ] }
 serde_json = "1"
-tokio = { version = "1.2", features = [ "time" ] }
+tokio = { version = "1", features = [ "time" ] }
 url = "2.2"
 vec1 = "1.6"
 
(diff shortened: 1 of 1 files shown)


clintfred (approved):
LGTM

The only improvement I can see is that I think for some post 1.0 libs, we probably don't need to specify the patch version in most case (anyhow, publicsuffix, vec1, url)

giarc3:
@clintfred please look again now that I removed some minimum patch versions (but only for post-1.0 crates)

clintfred (approved):
This looks good. Hopefully this will help decrease downstream compile times/bloat.

PR #226, opened 2021-02-25
Expand CI matrix and upload Cargo.lock

giarc3:


.github/workflows/ci.yaml
```diff
@@ -55,12 +56,16 @@ jobs:
...
-      - name: Cargo test (with one retry on failure)
-        uses: nick-invision/retry@v2.4.0
+      # Generate the Cargo.lock that will be used for the tests so we can upload it as an artifact
+      - name: Generate lockfile
+        run: cargo generate-lockfile
+      - name: Upload Cargo.lock
+        uses: actions/upload-artifact@v2
         with:
-          timeout_minutes: 30
-          max_attempts: 2
```
BobWall23, line 62: Don't need the timeout or max attempts any more?

BobWall23 (approved):

.github/workflows/ci.yaml
giarc3: We don't need to retry at all anymore now that we fixed ironcore-id

.github/workflows/ci.yaml
```diff
@@ -15,25 +15,26 @@ jobs:
       fail-fast: false
       matrix:
         # different operating systems
-        os: [ubuntu-18.04, macos-10.15, windows-2019]
-        features: [--features default]
+        os: [ubuntu-18.04, ubuntu-20.04, macos-10.15, windows-2019]
```
clintfred, line 18: maybe we should start migrating off 18.04?

.github/workflows/ci.yaml
```diff
@@ -15,25 +15,26 @@ jobs:
...
           - os: ubuntu-18.04
-            features: --features tls-rustls --no-default-features
+            features: tls-rustls --no-default-features
           - os: ubuntu-18.04
-            features: --features tls-vendored-openssl --no-default-features
+            features: tls-vendored-openssl --no-default-features
           # MSRV
           - os: ubuntu-18.04
             toolchain: 1.41.1
+            features: default
```
clintfred, line 37: this is redundant with `features: [default]` above, no?

giarc3: No, the above is a matrix that just changes the OS and has set features/toolchain. Down here are individual combinations added to the matrix, so they don't inherit the `features: [default]` from above. So without this line, the `cargo test` line gets made because it has `--features` with nothing specified.

.github/workflows/ci.yaml
giarc3: I would agree, but they're still using 18.04 as `ubuntu-latest`, so I would say we should wait until they change that before we use 20.04 as our primary OS in the matrix.

clintfred (approved):

PR #230, opened 2021-03-04
CI periodic testing

giarc3:
Run a simple test daily, the full matrix weekly on Mondays. Both happen at 2pm UTC, which should be 7 or 8am Mountain time depending on daylight savings. 
Sends an alert to the githubbery slack channel if the daily test fails. Also (hopefully) alerts if the weekly run fails, but I was having a hard time making sure it didn't also alert on a simple PR CI failure.
Also made some changes to the security audit to make it cleaner.
First attempt at addressing https://github.com/IronCoreLabs/design-docs/issues/584


cjyar (approved):

giarc3:
Issue in test is https://github.com/tokio-rs/mio/issues/1467

BobWall23 (approved):

coltfred:
Nice work!

coltfred (approved):

skeet70 (approved):

.github/workflows/daily.yaml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      # Generate the Cargo.lock that will be used for the tests so we can upload it as an artifact
+      - run: cargo generate-lockfile
+      - uses: actions/upload-artifact@v2
+        with:
+          path: Cargo.lock
+      - name: Decrypt test keys
+        run: openssl aes-256-cbc -d -K ${{ secrets.OPENSSL_KEY }} -iv ${{ secrets.OPENSSL_IV }} -in tests/testkeys/iak-dev.pem.enc -out tests/testkeys/iak-dev.pem
+      - run: cargo test --all-features
```
clintfred, line 32: I don't think this is probably what we want. I think we probably want `default` or just `default + beta` for this daily test.

clintfred (changes requested):

.github/workflows/daily.yaml
giarc3: I went with default + beta

clintfred (approved):

Issue #231, opened 2021-03-14
publicsuffix can be dropped


rushmorem:
It seems like this crate is only using `publicsuffix` to take advantage of `publicsuffix::IntoUrl` trait. This trait is no longer available in v2 of the crate. You can use `url::Url::parse` instead and drop  publicsuffix entirely.

clintfred:
@rushmorem Thanks for the heads up. Direct usage of `url` works great.

rushmorem:
You are welcome :)

PR #232, opened 2021-03-15
Remove dependency on publicsuffix

giarc3:
As discovered in #231, we were only using publicsuffix for the `IntoUrl` trait. Instead, we can just call `url::Url::parse` directly ourselves.


clintfred (approved):
LGTM

BobWall23 (approved):

Issue #241, opened 2021-07-19
Determine what might be limiting # of parallel decrypts 


BobWall23:
During TSP testing, it appears that ironoxide may be limiting the number of decrypt operations (which include a call to ironcore_id to fetch the transformed EDEK) that are being done in parallel.

Need to determine if there is something that is limiting the parallelism, or if it is something in the ironoxide consumer in the TSP.

skeet70:
AES decrypts and Recrypt transforms were both happening in the decrypt future after the network `DocumentGet` for metadata comes back. This meant that the future was doing a big chunk of CPU heavy crypto work without yielding or doing anything special to accommodate it. We should have a PR soon that uses `tokio::spawn_blocking` to explicitly move that section of the future onto a threadpool for heavy stuff, allowing the normal threadpool to keep cranking on the network work. This immediately improved CPU utilization.

PR #243, opened 2021-07-20
Use spawn_blocking to increase request throughput

giarc3:
Description in https://github.com/ironcorelabs/ironoxide/issues/241#issuecomment-883502432 by @skeet70:
> AES decrypts and Recrypt transforms were both happening in the decrypt future after the network DocumentGet for metadata comes back. This meant that the future was doing a big chunk of CPU heavy crypto work without yielding or doing anything special to accommodate it. We should have a PR soon that uses tokio::spawn_blocking to explicitly move that section of the future onto a threadpool for heavy stuff, allowing the normal threadpool to keep cranking on the network work. This immediately improved CPU utilization.


skeet70 (approved):

src/internal.rs
```diff
@@ -195,6 +195,9 @@ quick_error! {
         OperationTimedOut{operation: SdkOperation, duration: std::time::Duration} {
             display("Operation {} timed out after {}ms", operation, duration.as_millis())
         }
+        JoinError(err: tokio::task::JoinError) {
```
clintfred, line 198: I think the addition of this error type means we should do a minor version bump.

src/internal/document_api.rs
```diff
@@ -1055,31 +1055,35 @@ pub async fn document_update_bytes<
...
     let doc_meta = document_get_metadata(auth, &doc_header.document_id).await?;
-    let sym_key = transform::decrypt_as_symmetric_key(
-        recrypt,
-        doc_meta.0.encrypted_symmetric_key.clone().try_into()?,
-        device_private_key.recrypt_key(),
-    )?;
+    let device_private_key = device_private_key.clone();
+    tokio::task::spawn_blocking(move || {
+        let sym_key = transform::decrypt_as_symmetric_key(
+            &(*recrypt.clone()),
```
clintfred, line 1069:
```suggestion
            &recrypt,
```

clintfred (approved):
LGTM.

src/internal.rs
skeet70: https://doc.rust-lang.org/cargo/reference/semver.html#enum-variant-new looks like major is actually the recommendation for adding more variants to an enum, should probably do that.

giarc3: That's a cool page @skeet70, hadn't seen that before. It makes sense that it's considered a breaking change because someone matching over it would need to include the new type for the match to be exhaustive. Should we do a major bump or work to avoid that? If someone has a recommendation for an existing variant that works, I could use that.

If we do bump major, we could consider adding `#[non_exhaustive]` onto it so that we're free to add new variants in the future without a major bump.

clintfred: I was also wondering about `non_exhaustive`. I also realized, looking at this, that we would be adding a tokio error to our public API. We already expose error from ring and protobuf, but maybe we should stop doing that.

The chances of a consumer enumerating all of these error types in a `match` is effectively zero, so maybe allowing non-exhaustive would be fine.

skeet70: https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute for reference. In TSP code recently I wrote that `matches!` that enumerates the couple that mattered to me there and ignored the rest. AFAIK `non_exhaustive` would just always force people to do _something_ with the "rest" even if they've enumerated everything that exists right now. I don't think that would've affected my code. It could be a minor annoyance if you really did want to be exhaustive (and know you were) but I can't think of a use case where you wouldn't be able to have some catch all fallback at the application level. 

https://rust-lang.github.io/rfcs/2008-non-exhaustive.html#enums the RFC for it explicitly called out error types being the most common use case.

I think we make this `non_exhaustive`.

clintfred (approved):
Merge on green and release, please.

PR #246, opened 2021-11-04
Don't clone plaintext on aes encryption

### LINKED ISSUE #3: Avoid cloning plaintext on AES encrypt
In our AES encrypt method in crypto/aes.rs, we're cloning the provided plaintext vector so we can resize it to add 16 bytes to it to support adding on the GCM auth tag. As people could be encrypting large large documents with the SDK, doing a clone is bad. We need to figure out a way to avoid it.

It would also be nice to see if we can have the various AES methods take slices instead of vecs as some inputs to encrypt have to be copied to get them into vectors. See [this code review comment](https://github.com/IronCoreLabs/ironrust/pull/75/files#r267116017)

giarc3:
PR really has 2 parts, so I broke them into 2 commits you can view separately. I think we should definitely keep 1, but we could decide to not do 2.
1. `crypto::aes::encrypt` now takes `Vec<u8>` plaintext instead of `&[u8]`. The call sites were all already doing `&plaintext.to_vec()`, so they had owned data. And now we don't have to `.to_owned()` when encrypting.
2. Changes our public functions like `document_encrypt()` to take `Vec<u8>` instead of `&[u8]`. Since this is public this would be a breaking change. But essentially if we need owned data, we should take owned data (instead of just internally cloning with `.to_vec()`). It'll instead be up to the caller whether or not they need to clone their data (and I think they frequently won't need to be using their unencrypted_data again after encrypting it).

Closes #3

### DIFF
--- a/examples/encrypting.rs
+++ b/examples/encrypting.rs
@@ -18,10 +18,10 @@ async fn main() -> Result<()> {
 async fn encrypt_to_group(sdk: &IronOxide) -> Result<DocumentId> {
     // start-snippet{encryptToGroup}
     let group_id = create_group(sdk).await?;
-    let message = "This is my secret which a whole group should see.";
+    let message = "This is my secret which a whole group should see.".to_string();
     let encrypted_result = sdk
         .document_encrypt(
-            message.as_bytes(),
+            message.into_bytes(),
             &DocumentEncryptOpts::with_explicit_grants(None, None, true, vec![(&group_id).into()]),
         )
         .await?;
@@ -32,10 +32,10 @@ async fn encrypt_to_group(sdk: &IronOxide) -> Result<DocumentId> {
 
 async fn encrypt_to_user(sdk: &IronOxide, user_id: &UserId) -> Result<DocumentId> {
     // start-snippet{encryptToUser}
-    let message = "This is my secret for a single user.";
+    let message = "This is my secret for a single user.".to_string();
     let encrypted_result = sdk
         .document_encrypt(
-            message.as_bytes(),
+            message.into_bytes(),
             &DocumentEncryptOpts::with_explicit_grants(None, None, true, vec![user_id.into()]),
         )
         .await?;
@@ -47,7 +47,7 @@ async fn encrypt_to_user(sdk: &IronOxide, user_id: &UserId) -> Result<DocumentId
 
 async fn encrypt_with_policy(sdk: &IronOxide) -> Result<DocumentId> {
     // start-snippet{encrypt_with_policy}
-    let message = "this is my secret which has some labels.";
+    let message = "this is my secret which has some labels.".to_string();
     let data_labels = PolicyGrant::new(
         Some(Category::try_from("PII")?),
         Some(Sensitivity::try_from("PRIVATE")?),
@@ -56,7 +56,7 @@ async fn encrypt_with_policy(sdk: &IronOxide) -> Result<DocumentId> {
     );
     let encrypted_result = sdk
         .document_encrypt(
-            message.as_bytes(),
+            message.into_bytes(),
             &DocumentEncryptOpts::with_policy_grants(None, None, data_labels),
         )
         .await?;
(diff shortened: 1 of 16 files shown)


coltfred (commented):
This is an improvement for sure because we were cloning the Vec and then the underlying aes library extends that vec, which clones it again.

If we're planning to break the public api maybe we should go further and use a similar type to what ring uses?

```
        InOut: AsMut<[u8]> + for<'in_out> Extend<&'in_out u8>,
```
Fixing things to `Vec<u8>` means that people using other byte representations must convert them to vecs first.

Did you give any thought to this?

BobWall23:
It looks like the change necessitated a lot of `clone` calls in code that was encrypting the data - so maybe it didn't remove the problem, it just moved it up to the consumer?

In some of the tests, instead of cloning the vector, the code could have just grabbed the length before calling encrypt, so maybe the clones could be avoided in more places.

examples/encrypting.rs
```diff
@@ -18,10 +18,10 @@ async fn main() -> Result<()> {
 async fn encrypt_to_group(sdk: &IronOxide) -> Result<DocumentId> {
     // start-snippet{encryptToGroup}
     let group_id = create_group(sdk).await?;
-    let message = "This is my secret which a whole group should see.";
+    let message = "This is my secret which a whole group should see.".to_string();
     let encrypted_result = sdk
         .document_encrypt(
-            message.as_bytes(),
+            message.into_bytes(),
```
clintfred, line 24: I somehow have a slight preference for: 
```rust
let message = "This is my secret which a whole group should see.";
// ...
message.as_bytes().to_vec();
// or
message.as_bytes.to_owned();
```

src/document.rs
```diff
@@ -154,14 +154,14 @@ pub trait DocumentOps {
     /// # use ironoxide::prelude::*;
     /// # let sdk: IronOxide = unimplemented!();
     /// # use ironoxide::document::DocumentEncryptOpts;
-    /// let data = "secret data".as_bytes();
+    /// let data = "secret data".to_string().into_bytes();
```
clintfred, line 157: again, I think `"secret data".as_bytes().to_owned()` might be slightly better. Most important is consistency in our examples, but just pointing this one out in case the other one changes.

clintfred (changes requested):
Interesting PR.

I think if we do this, we should probably just provide it as an additive change and call the owned variants `*_owned` or `*_vec`

From a memory size perspective there is certainly a case to being able to chose a "faster" path if you know you don't need the data any more.

The main downside I see is that we are telling them the _kind_ of owned data they must give us. Furthermore, we are specifying that it must be on the heap. There are other byte representations that use arrays underneath that would have to convert to Vec first.  (arrayvec, smallvec, etc)

I like @coltfred's suggestion of seeing if we can be exactly restrictive as the underlying library with something like 
```
InOut: AsMut<[u8]> + for<'in_out> Extend<&'in_out u8>,
```

Could we encrypt right on the stack then, for example?

examples/encrypting.rs
giarc3: I think I'd prefer to leave this as it is. I agree that in this case with a contrived `&str` it might make more sense using `as_bytes()`, but I think it's just as likely that consumers will have an owned `String` message instead, in which case this is how they'd turn it into bytes.
Let me know if you'd like to discuss this or have me think on it further.

giarc3:
After a few attempts to make the encrypt/decrypt functions more generic (`InOut: AsMut<[u8]> + for<'in_out> Extend<&'in_out u8>`), I was not able to make this work (especially not in an async context). However, I think that this PR is still an improvement as it stands now and is still worth merging. 
Previously on encrypt, we would take a borrowed byte slice, clone it into a Vec, then clone it again when `encrypt` extends the Vec. 
Now we take a Vec, and still have the clone due to the extend.
If the consumer already has a Vec, we saved a clone of the plaintext.
If the consumer has a byte slice or other byte representation, they will have to clone it themselves before passing it in, and we're in the same place as when we started. No better, no worse.

@coltfred @clintfred Please take another look and see if you agree with my assessment.

coltfred:
I agree with your assessment. It is better and we can continue to make it better if we find a way to make the api more ergonomic in the future. For now it's an improvement where we use it (and will be for most people I suspect).

coltfred (approved):

clintfred (approved):
I agree this is a step in the right direction.
We should create a changelog entry for this before merging.

PR #247, opened 2021-11-05
Satisfy clippy

giarc3:
A few lints were just silenced: too many arguments, complex return type, and large enum size difference.
This did lead to some changes in our public API, so we'll have to decide if we're okay with that (changing anywhere we take `&Vec<_>` to `&[_]`).
With this satisfied, we could consider adding the clippy workflow to this repo's CI

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -147,10 +147,12 @@
 //! # }
 //! ```
 
-// required by quick_error or IronOxideErr
-#![recursion_limit = "128"]
-// required as of rust 1.46.0
-#![type_length_limit = "2000000"]
+#![allow(clippy::too_many_arguments)]
+#![allow(clippy::type_complexity)]
+// // required by quick_error or IronOxideErr
+// #![recursion_limit = "128"]
+// // required as of rust 1.46.0
+// #![type_length_limit = "2000000"]
 
 // include generated proto code as a proto module
 include!(concat!(env!("OUT_DIR"), "/transform.rs"));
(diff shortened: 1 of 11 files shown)


BobWall23 (approved):

src/lib.rs
```diff
@@ -147,9 +147,11 @@
 //! # }
 //! ```
 
-// required by quick_error or IronOxideErr
+#![allow(clippy::too_many_arguments)]
+#![allow(clippy::type_complexity)]
+// // required by quick_error or IronOxideErr
```
clintfred, line 152: side-note: I wonder if this is still required?
```suggestion
// required by quick_error or IronOxideErr
```

src/lib.rs
```diff
@@ -147,9 +147,11 @@
...
 #![recursion_limit = "128"]
-// required as of rust 1.46.0
+// // required as of rust 1.46.0
```
clintfred, line 154:
```suggestion
// required as of rust 1.46.0
```

clintfred (approved):
The main change here is `&Vec<T> -> &[T]` which should be completely backwards compatible. 
There is a couple of commenting snafus snafus, but otherwise all good changes.

src/lib.rs
giarc3: I don't think they are necessary, which is why I was commenting them out and messed up the un-comment. But I didn't want to fully test it as part of this PR.

PR #248, opened 2021-11-19
Rust 2021, dependency upgrades

giarc3:
- Update to recrypt 0.13
  - Update to rand 0.8
  - Update to rand_chacha 0.3
  - Update to ironcore-search-helpers 0.2

### DIFF
--- a/src/blocking.rs
+++ b/src/blocking.rs
@@ -273,7 +273,7 @@ impl BlockingIronOxide {
 
 /// Creates a tokio runtime with the default number of core threads (num of cores on a machine)
 fn create_runtime() -> tokio::runtime::Runtime {
-    tokio::runtime::Builder::new_multi_thread()
+    tokio::runtime::Builder::new_current_thread()
         .enable_all() // enable both I/O and time drivers
         .build()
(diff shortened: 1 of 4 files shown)


coltfred (approved):
Finally getting rid of the crusty rand. Nice!

Also great that the new resolver for 2021 found that bug in the blocking api.

src/blocking.rs
```diff
@@ -273,7 +273,7 @@ impl BlockingIronOxide {
 
 /// Creates a tokio runtime with the default number of core threads (num of cores on a machine)
 fn create_runtime() -> tokio::runtime::Runtime {
-    tokio::runtime::Builder::new_multi_thread()
+    tokio::runtime::Builder::new_current_thread()
```
clintfred, line 276: Given that there's no upside to having multiple threads since all the calls use `block_on`, this is a good change. You should probably update the comment above.

clintfred (approved):

src/blocking.rs
giarc3: :+1:

PR #249, opened 2021-11-19
Removes `chrono` and  usages  of `DateTime<Utc>` and replaces them with`time::OffsetDateTime`

### LINKED ISSUE #244: RUSTSEC-2020-0071: Potential segfault in the time crate
> Potential segfault in the time crate

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `time`                      |
| Version             | `0.1.43`                   |
| URL                 | [https://github.com/time-rs/time/issues/293](https://github.com/time-rs/time/issues/293) |
| Date                | 2020-11-18                         |
| Patched versions    | `>=0.2.23`                  |
| Unaffected versions | `=0.2.0,=0.2.1,=0.2.2,=0.2.3,=0.2.4,=0.2.5,=0.2.6`               |

### Impact

Unix-like operating systems may segfault due to dereferencing a dangling pointer in specific circumstances. This requires an environment variable to be set in a different thread than the affected functions. This may occur without the user&#39;s knowledge, notably in a third-party library.

The affected functions from time 0.2.7 through 0.2.22 are:

- `time::UtcOffset::local_offset_at`
- `time::UtcOffset::try_local_offset_at`
- `time::UtcOffset::current_local_offset`
- `time::UtcOffset::try_current_local_offset`
- `time::OffsetDateTime::now_local`
- `time::OffsetDateTime::try_now_local`

The affected functions in time 0.1 (all versions) are:

- `at`
- `at_utc`

Non-Unix targets (including Windows and wasm) are unaffected.

### Patches

Pending a proper fix, the internal method that determines the local offset has been modified to always return `None` on the affected operating systems. This has the effect of returning an `Err` on the `try_*` methods and `UTC` on the non-`try_*` methods.

Users and library authors with time in their dependency tree should perform `cargo update`, which will pull in the updated, unaffected code.

Users of time 0.1 do not have a patch and should upgrade to an unaffected version: time 0.2.23 or greater or the 0.3. series.

### Workarounds

No workarounds are known.

### References

time-rs/time#293

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2020-0071.html) for additional details.

clintfred:
Closes #244 #245 

This change should not have any semantic meaning and is being done to address RUSTSEC-2020-0071 (see #244) and RUSTSEC-20200159 (see #245).

This is a breaking API change because some result types include create/updated timestamps. Consumers of the library will need to add a direct dependency on `time`. Note that `chrono` already uses `time` internally, so this change should not bloat transitive dependencies.

For those needing serialization support of  `OffsetDateTime` timestamps, the `time` crate offers this support with the `serde` feature flag.

One small point of friction here is rfc3339 support. `time` has this support on [a branch](https://github.com/time-rs/time/issues/387) that is blocked on a rust feature hitting stable, so I copied in a few lines of code from that branch to support deserializing rfc3339 timestamps.

To fully fix the RUSTSEC advisories, it was necessary to upgrade one  of our dependencies as well. 
See https://github.com/Keats/jsonwebtoken/pull/222

For now, this PR depends on https://github.com/IronCoreLabs/jsonwebtoken.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.27.0 (Unreleased)
+- [[#249](https://github.com/IronCoreLabs/ironoxide/pull/249)] Remove `chrono` types in public API and replace with equivalent `time` types.
+
 ## 0.26.0
 
 - [[#243](https://github.com/IronCoreLabs/ironoxide/pull/243)] Add `#[non_exhaustive]` to IronOxideErr.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,8 @@ futures = "0.3.1"
 hex = "0.4"
 ironcore-search-helpers = { version = "0.1.2", optional = true }
 itertools = "0.10"
-jsonwebtoken = { git = "https://github.com/IronCoreLabs/jsonwebtoken" }
+# see https://github.com/IronCoreLabs/ironoxide/pull/249
+jsonwebtoken = { git = "https://github.com/IronCoreLabs/jsonwebtoken", rev = "393782" }
 lazy_static = "1.4"
 log = "0.4"
 percent-encoding = "2.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ hex = "0.4"
 ironcore-search-helpers = { version = "0.2", optional = true }
 itertools = "0.10"
 # see https://github.com/IronCoreLabs/ironoxide/pull/249
-jsonwebtoken = { git = "https://github.com/IronCoreLabs/jsonwebtoken", rev = "393782" }
+jsonwebtoken = { git = "https://github.com/IronCoreLabs/jsonwebtoken", rev = "147e560" }
 lazy_static = "1.4"
 log = "0.4"
 percent-encoding = "2.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ironoxide"
-version = "0.26.0"
+version = "0.27.0"
 authors = [ "IronCore Labs <info@ironcorelabs.com>" ]
 readme = "README.md"
 license = "AGPL-3.0-only"
@@ -27,7 +27,7 @@ futures = "0.3.1"
 hex = "0.4"
 ironcore-search-helpers = { version = "0.1.2", optional = true }
 itertools = "0.10"
-jsonwebtoken = "7.2"
+jsonwebtoken = { git = "https://github.com/IronCoreLabs/jsonwebtoken" }
 lazy_static = "1.4"
 log = "0.4"
 percent-encoding = "2.1"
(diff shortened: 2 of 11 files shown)


Cargo.toml
```diff
@@ -27,7 +27,7 @@ futures = "0.3.1"
 hex = "0.4"
 ironcore-search-helpers = { version = "0.1.2", optional = true }
 itertools = "0.10"
-jsonwebtoken = "7.2"
+jsonwebtoken = { git = "https://github.com/IronCoreLabs/jsonwebtoken" }
```
giarc3, line 30: When we depend on forks we often lock it to a commit. Also, we could consider upgrading jsonwebtoken's dependency on base64 0.13 while we're here, since it's what we depend on.

src/internal/group_api/requests.rs
```diff
@@ -484,24 +490,22 @@ pub mod group_remove_entity {
...
-
     use super::*;
     use recrypt::api::KeyGenOps;
 
     ///This test is to ensure our lowercase admin and member permissions are handled correctly.
     #[test]
     fn group_item_serde_format_is_expected() {
-        let created = Utc.timestamp_millis(1_551_461_529_000);
-        let updated = Utc.timestamp_millis(1_551_461_529_001);
+        let created = OffsetDateTime::from_unix_timestamp_nanos(1_551_461_529_000_000).unwrap();
```
clintfred, line 507: This isn't right. This is microseconds. Need to investigate.

src/internal/serde_rfc3339.rs
```diff
@@ -0,0 +1,54 @@
+//! MAINTENANCE NOTE
+//! This code is copied from https://github.com/time-rs/time/tree/serde-rfc3339
```
clintfred, line 2: Should insert permalink here.

src/internal/group_api/requests.rs
clintfred: The test didn't really care about this number... I did increase it by another factor of 1000.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.27.0 (Unreleased)
```
BobWall23, line 3: Let's start putting dates on the changelog entries.
Also, any reason to not release this?

BobWall23 (changes requested):

CHANGELOG.md
clintfred: Not necessarily disagreeing on adding the changelog dates, but the canonical information for release-dates is available from crates.io via
```
curl 'https://crates.io/api/v1/crates/ironoxide' 
```

As for release - I think it's fine to release this whenever, but we should merge this PR first and then follow https://github.com/IronCoreLabs/ironoxide/blob/main/RELEASING.md 
If we want to start adding dates in CHANGELOG.md, adding it to the release checklist would be the right place.

coltfred (approved):
I think this looks good to me. I did bump the base64 dep to 0.13 in jsonwebtoken to match ours. That's what they're doing on v8 of the real release.

PR #258, opened 2022-02-17
Update to v8 of jsonwebtoken

coltfred:


src/internal/user_api.rs
```diff
@@ -309,10 +309,17 @@ impl Jwt {
     ///
     /// Verifies that the provided jwt uses a compatible algorithm and contains the required claims.
     pub fn new(jwt: &str) -> Result<Jwt, IronOxideErr> {
-        // This `dangerous_unsafe_decode` is safe because the server will do the actual
+        let bogus_key = DecodingKey::from_secret(&[]);
+        let validation = {
+            let mut temp: Validation = Default::default();
+            temp.insecure_disable_signature_validation();
+            temp.validate_exp = false;
+            temp
```
giarc3, line 317: It looks like the Validation says 
>    /// The validation will check that the `alg` of the header is contained
>    /// in the ones provided and will error otherwise. Will error if it is empty.
>   ///
>  /// Defaults to `vec![Algorithm::HS256]`.

But don't we want the algorithm to be ES256 or HS256?

Also, the comment says "We just want to do a little initial validation to catch issues earlier." Are we doing that? What things are we validating to be true here?

PR #274, opened 2022-06-17
Make sure CI uses correct Rust toolchain

giarc3:


.github/workflows/ci.yaml
```diff
@@ -38,14 +38,15 @@ jobs:
             features: default
           # MSRV
           - os: ubuntu-18.04
-            toolchain: 1.56.0
+            toolchain: 1.60.0
             features: default
     steps:
       - uses: actions/checkout@v3
```
BobWall23, line 44: If we bump this to v4 and set the python version now, will that avoid a workflow update on Monday?

BobWall23 (approved):

coltfred (approved):

PR #281, opened 2022-09-09
Make JWT struct more lenient

giarc3:
Our JWT validation in ironoxide is intended to be a simple sanity check before sending to the server (which does actual validation). We were previously not allowing both `http://ironcore/sid` and `sid` claims to be present as serde would attempt to put them both in the `sid` field. The claims and their prefixed versions are now all Options, and we validate that at least one of each is present.


src/internal/user_api.rs
```diff
@@ -915,4 +953,14 @@ pub(crate) mod tests {
...
+        // { "http://ironcore1/pid": 1, "http://ironcore1/kid": 1859, "http://ironcore1/sid": "IronHide", "sid": "IronHide",
+        //    "http://ironcore1/uid": "bob.wall@ironcorelabs.com", "iss": "https://ironcorelabs.auth0.com/",
+        //    "sub": "github|11368122", "aud": "hGELxuBKD64ltS4VNaIy2mzVwtqgJa5f", "iat": 1593130255, "exp": 1593133855 }
+        let jwt = Jwt::try_from("eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJodHRwOi8vaXJvbmNvcmUvcGlkIjoxLCJodHRwOi8vaXJvbmNvcmUva2lkIjoxODU5LCJodHRwOi8vaXJvbmNvcmUvc2lkIjoiSXJvbkhpZGUiLCJzaWQiOiJJcm9uSGlkZSIsImh0dHA6Ly9pcm9uY29yZS91aWQiOiJib2Iud2FsbEBpcm9uY29yZWxhYnMuY29tIiwiaXNzIjoiaHR0cHM6Ly9pcm9uY29yZWxhYnMuYXV0aDAuY29tLyIsInN1YiI6ImdpdGh1YnwxMTM2ODEyMiIsImF1ZCI6ImhHRUx4dUJLRDY0bHRTNFZOYUl5Mm16Vnd0cWdKYTVmIiwiaWF0IjoxNTkzMTMwMjU1LCJleHAiOjE1OTMxMzM4NTV9.Bp5aXzT0k0busaETapSaOUyrEq1zkW-iLCyVEFP7ngEOJY1eU3BAuuDsIZmpW53oVCcCrgGt43VRse6oYegoz_ZsR2Zn2GjQ5JMPRdTZxrV5rjATJeCl9LZwwLktur0ik5j1KLDSA_hAlp8oaHfS50gtxh5qfG3e42o72jvdBcC9cOT45e2xEVPD67x7Ijt2NKPG7HQOyvKILMFuQtH1M0ywo9yNwJe4xS5n_B8F_PYZLe-EElSQY-ydXukfn0k1sQxt8VhgHZRUnKgmKVej5Z6_OtvgKONATIKJSnv1aI7C6il2yG7Di-b5YZr8R2hej57adBeISVoUdbOpgBON4g");
```
coltfred: I think we should test all the duplicates.

coltfred (approved):

src/internal/user_api.rs
giarc3: Done

skeet70 (approved):

PR #284, opened 2022-09-22
Test against staging, not dev.

cjyar:
Removes all traces of the dev IRONCORE_ENV and moves CI to test against stage. Also fixes an old bug with one of the CI matrix runs.

I've disabled the action that posts to Slack if this fails. We don't need it because other automation is already monitoring for workflow failures.


benches/ironoxide_bench.rs
```diff
@@ -6,17 +6,16 @@ use tokio::runtime::Runtime;
...
+    /// Reads from the environment variable IRONCORE_ENV. Supports `stage`, and `prod`.
+    /// An unset environment variable will be interpreted as `stage`.
     /// An invalid environment variable will result in a panic.
     pub static ref ENV: String = match std::env::var("IRONCORE_ENV") {
         Ok(url) => match url.to_lowercase().as_str() {
-            "dev" => "dev",
             "stage" => "stage",
             "prod" => "prod",
-            _ => panic!("IRONCORE_ENV can only be set to `dev`, `stage`, or `prod` when running the benchmarks.")
+            _ => panic!("IRONCORE_ENV can only be set to `stage`, or `prod` when running the benchmarks.")
```
cjyar, line 16:
```suggestion
            _ => panic!("IRONCORE_ENV can only be set to `stage` or `prod` when running the benchmarks.")
```

coltfred (approved):

PR #301, opened 2023-09-12
Don't use the Rust cache for daily builds.

cjyar:
Leaving the cache directive, commented out, with a comment to prevent somebody from re-adding it.


cjyar:
I can't test this, even though I added `workflow_dispatch` to the workflow. The `workflow_dispatch` has to exist on the workflow in the default branch before it'll let you invoke it on any other branch.

BobWall23 (approved):

giarc3 (approved):

PR #325, opened 2024-09-09
Make a lazy static reqwest client and use it

skeet70:


skeet70:
In benches run locally back to back this led to a modest (~ -30 to +1%) performance improvement, with pretty wide confidence intervals.

PR #339, opened 2025-02-26
Rust 2024 edition

giarc3:


src/internal.rs
```diff
@@ -295,8 +295,8 @@ pub mod auth_v2 {
         pub(in crate::internal::auth_v2) timestamp: OffsetDateTime,
     }
 
-    impl<'a> AuthV2Builder<'a> {
-        pub fn new(req_auth: &'a RequestAuth, timestamp: OffsetDateTime) -> AuthV2Builder {
+    impl AuthV2Builder<'_> {
```
BobWall23, line 298: This seems suspicious.

src/internal.rs
```diff
@@ -805,7 +805,7 @@ pub(crate) mod tests {
     use vec1::vec1;
 
     /// String contains matcher to assert that the provided substring exists in the provided value
-    pub fn contains<'a>(expected: &'a str) -> Box<dyn Matcher<String> + 'a> {
+    pub fn contains(expected: &str) -> Box<dyn Matcher<String> + '_> {
```
BobWall23, line 808: Also weird - does it now infer the lifetime somehow?

BobWall23 (approved):

src/internal.rs
giarc3: Just a clippy lint about eliding the lifetime

src/internal.rs
giarc3: Same

PR #365, opened 2026-02-27
Add offline encryption support via public key cache

### LINKED ISSUE #359: Support offline encryption
Initial thoughts were to cache some set of transforms/public keys, and used those cached values for encryption.

skeet70:
Closes #359 

I think this aligns with everything we laid out in the ticket and discussed. There are a few added `TODO(breaking)` things that we'd like to do when we're breaking for some other reason, such as removing the concept of owners. I tried to add tests for everything I could come up with, then added more that AI could come up with. I ran the benchmarks to confirm nothing was crazy there. The core of the changes is in `src/internal.rs` and `src/lib.rs`, and it propagates up via `internal::user_api::get_user_keys` and `internal::group_api::get_group_keys` which both now generically call `internal::get_keys_with_cache` and are called by `internal::document_api::resolve_keys_for_grants`.

I also modified the example app to run unmanaged encrypts in a loop and timed it while printing cache hits, it works as expected (first call is ~110ms, rest are ~5ms).

This also fully fleshes out our `_unmanaged` calls, even in places where they're the same (to simplify usage for someone trying to do unmanaged). Those changes are all in `src/document/advanced.rs`.


src/internal.rs
```diff
@@ -426,6 +432,128 @@ impl DeviceContext {
...
+            .partition_map(|id| match ids_with_keys.get(&id).cloned() {
+                Some(pub_key) => {
+                    cache.pin().insert(id.clone(), pub_key.clone());
+                    Either::Right(WithKey::new(id, pub_key))
+                }
+                None => Either::Left(id),
+            });
+
+    // cache the newly returned API values
+    for id_with_key in &found {
```
coltfred: You're already inserting these as a side effect in the partition map. I'd add a comment up there and remove this one.

src/blocking.rs
```diff
@@ -169,7 +169,7 @@ impl BlockingIronOxide {
         self.runtime
             .block_on(self.ironoxide.document_revoke_access(id, revoke_list))
     }
-    /// See [ironoxide::document::advanced::DocumentAdvancedOps::document_encrypt_unmanaged](trait.DocumentAdvancedOps.html#tymethod.document_encrypt_unmanaged)
+    /// See [ironoxide::document::advanced::DocumentAdvancedOps::document_encrypt](trait.DocumentAdvancedOps.html#tymethod.document_encrypt_unmanaged)
```
coltfred: You changed the text, but not the place it links to.

src/internal.rs
```diff
@@ -426,6 +432,128 @@ impl DeviceContext {
...
+    user_keys: UserPublicKeyCache,
+    #[serde(serialize_with = "serialize_papaya_map")]
+    group_keys: GroupPublicKeyCache,
+}
+
+/// `papaya` can't provide `size_hint` (due to being concurrent, no reliable upper bound) to indicate to serde's
+/// `serialize_seq` how long it is during serialization. `postcard` (our data format) requires a length on
+/// `serialize_seq` or it errors. This manual serialization gets around `papaya`'s lack of `size_hint` by serializing
+/// individual entries till it runs out, which dodges some concurrency issues (something already serialized could get
+/// tombstoned while writing and be out of date).
```
coltfred: This comment is a lot of caveat that we can't reliably test. Seems worth it to just pin, collect into a Vec and provide serialization we know will work. Even a fairly large cache will still be relatively small in terms of overall memory.

src/internal.rs
```diff
@@ -1335,4 +1479,313 @@ pub(crate) mod tests {
...
+            Ok(io)
+        }
+        #[test]
+        fn signed_cache_roundtrip() -> Result<()> {
+            let io = create_test_sdk()?;
+            let recrypt = Recrypt::new();
+            let device = super::create_test_device_context();
+            let (_, pubk) = recrypt.generate_key_pair()?;
+
+            let cache = PublicKeyCache::default();
```
coltfred: You're creating a local cache here, but using `io` for the test below.

src/document/advanced.rs
```diff
@@ -51,6 +68,132 @@ pub trait DocumentAdvancedOps {
...
+    /// let users_or_groups: Vec<UserOrGroup> = users.iter().map(|user| user.into()).collect();
+    /// let access_result = sdk.document_grant_access_unmanaged(&edeks, &users_or_groups).await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    fn document_grant_access_unmanaged(
+        &self,
+        edeks: &[u8],
+        grant_list: &[UserOrGroup],
+    ) -> impl Future<Output = Result<DocumentAccessResult>> + Send;
```
coltfred: This return type isn't correct. The return type in the case of unmanaged will have to be the successful and failed ids and the updated edeks (which will just be bytes). 

You could probably get away with just errors and the updated edeks, but if we're going to mirror the `DocumentEncryptUnmanagedResult` we should include the grants as well.

src/document/advanced.rs
```diff
@@ -51,6 +68,132 @@ pub trait DocumentAdvancedOps {
...
+    /// let users_or_groups: Vec<UserOrGroup> = users.iter().map(|user| user.into()).collect();
+    /// let access_result = sdk.document_revoke_access_unmanaged(&edeks, &users_or_groups).await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    fn document_revoke_access_unmanaged(
+        &self,
+        edeks: &[u8],
+        revoke_list: &[UserOrGroup],
+    ) -> impl Future<Output = Result<DocumentAccessResult>> + Send;
```
coltfred: Ditto to the above, since these are unmanaged we just need to remove the edeks and return the updated edeks value.

src/document/advanced.rs
```diff
@@ -115,4 +259,72 @@ impl DocumentAdvancedOps for crate::IronOxide {
...
+    fn document_get_id_from_edeks_unmanaged(&self, edeks: &[u8]) -> Result<DocumentId> {
+        let proto_edeks = EncryptedDeksP::parse_from_bytes(edeks).map_err(IronOxideErr::from)?;
+        (*proto_edeks.documentId).try_into()
+    }
+
+    async fn document_grant_access_unmanaged(
+        &self,
+        edeks: &[u8],
+        grant_list: &[UserOrGroup],
+    ) -> Result<DocumentAccessResult> {
```
coltfred: For granting access to unmanaged documents we need to use a similar method to what's being done in [encrypt_document_unmanaged](https://github.com/IronCoreLabs/ironoxide/blob/1013ebdf2475dd45221b0a70c9d6538b7d6c3967/src/internal/document_api.rs#L783). In the context of grant, we'll have to do the following:

1 - Transform the edek using the ironcore service `edek_transform` is the call we'll make.
2 - Decrypt that value into a `Plaintext`.
3 - Encrypt that `Plaintext` to each of the grants
4 - combine all the edeks together and return them 

I'm happy to help with this as some of it will not be code that ironoxide has exposed before. All the functionality exists, but might need to be broken out from existing functions.

src/document/advanced.rs
```diff
@@ -115,4 +259,72 @@ impl DocumentAdvancedOps for crate::IronOxide {
...
+            SdkOperation::DocumentGrantAccess,
+        )
+        .await?
+    }
+
+    async fn document_revoke_access_unmanaged(
+        &self,
+        edeks: &[u8],
+        revoke_list: &[UserOrGroup],
+    ) -> Result<DocumentAccessResult> {
```
coltfred: Like above, this will need to return the edeks. Luckily it's much simpler as you just iterate over the edeks and delete the ones that are associated with the users/groups.

coltfred (changes requested):
I have to run away from the computer for a while so I'll submit my start at a review. We should also add integration tests for these operations in the document_ops file.

src/internal.rs
skeet70: There's no caveat, that's all explaining _why_ this serialize exists, all the caveats/weirdness applies to the default one in papaya. We effectively pin an approximate upper bound here with the `Some(map.len())` and start iterating, serializing as we go. If something is concurrently removed ahead of us in the iterator it won't be printed and the length we provided will be wrong, but that won't error anything. If things are concurrently added past our length, they won't be printed in this serialize call (same as collecting to a vec first).

Edit: I'll Vec of entries for ser/de custom.

src/document/advanced.rs
```diff
@@ -115,4 +262,64 @@ impl DocumentAdvancedOps for crate::IronOxide {
         )
         .await?
     }
+
+    async fn document_get_metadata_unmanaged(
+        &self,
+        edeks: &[u8],
+    ) -> Result<DocumentMetadataResult> {
+        add_optional_timeout(
+            document_api::document_get_metadata(
```
coltfred: Since the document is unmanaged the metadata for this cannot be retrieved from ironcore-id, it must be entirely processed from the edeks. Also we should have an integration test on this function.

src/internal/document_api.rs
```diff
@@ -1126,6 +1197,159 @@ pub async fn decrypt_document_unmanaged<CR: rand::CryptoRng + rand::RngCore>(
...
+        .iter()
+        // a failure to parse from proto here means it won't show up in access failures, but we have no ID to provide
+        // a failure for in this case anyway.
+        .filter_map(|edek| edek.userOrGroup.as_ref()?.UserOrGroupId.as_ref())
+        .filter_map(|proto_user_or_group| match proto_user_or_group {
+            UserOrGroupIdP::UserId(u) => UserId::try_from(u.to_string()).ok().map(Either::Left),
+            UserOrGroupIdP::GroupId(g) => GroupId::try_from(g.to_string()).ok().map(Either::Right),
+        })
+        .partition_map(identity);
+    let user_grants: Vec<_> = [existing_users.as_slice(), users]
```
coltfred: We shouldn't encrypt again to people they didn't ask us to. The existing might be very large and they're just encrypting to some. We should encrypt to all they users and groups they asked us to though (replacing the edeks from the old proto).

src/internal/document_api.rs
```diff
@@ -1982,4 +2205,270 @@ mod tests {
 
         Ok(())
     }
+
+    /// Helper: create EDEK proto bytes with the given user/group grants for a document.
+    fn make_edek_bytes(
```
coltfred: I think I'd name this something that implies that this creates a new dek to encrypt and encrypts it to the users/groups would be good. Maybe even `build_test_edek_bytes` or something.

You can also simplify the end of this function so you don't have to make up an AesEncryptedValue.
```

        let document_id = DocumentId(doc_id.into());
        let recryption_result = recrypt_document(
            &signingkeys,
            &recr,
            recr.gen_plaintext(),
            &document_id,
            with_keys,
        )?;

        edeks_to_edeks_proto(&recryption_result.edeks, doc_id, seg_id)
    }
```

src/internal.rs
```diff
@@ -426,6 +432,134 @@ impl DeviceContext {
...
+    pub(crate) fn deserialize(serialized_cache: &[u8]) -> Result<Self> {
+        postcard::from_bytes(serialized_cache).map_err(IronOxideErr::CacheSerdeError)
+    }
+    pub(crate) fn user_keys(&self) -> &HashMap<UserId, PublicKey> {
+        &self.user_keys
+    }
+    pub(crate) fn group_keys(&self) -> &HashMap<GroupId, PublicKey> {
+        &self.group_keys
+    }
+    pub(crate) fn deserialize_signed_public_key_cache(
```
coltfred, line 499: I was expecting to see a serialize function right next to this, but the serialize is inlined inside lib. We should move it down here so it's obvious to future us the way it's created and consumed.

src/internal.rs
```diff
@@ -426,6 +432,134 @@ impl DeviceContext {
...
+fn deserialize_papaya_map<'de, D, K, V>(deserializer: D) -> StdResult<HashMap<K, V>, D::Error>
+where
+    D: Deserializer<'de>,
+    K: Deserialize<'de> + core::hash::Hash + Eq,
+    V: Deserialize<'de>,
+{
+    let entries: Vec<(K, V)> = Vec::deserialize(deserializer)?;
+    let map = HashMap::new();
+    for (k, v) in entries {
+        map.pin().insert(k, v);
```
coltfred: Nitpick: You can pin once outside the loop.

coltfred (changes requested):
Couple remaining things. I also added some integration tests. Some of them have TODOs because they should behave differently and 1 that is broken on purpose and it should pass.

You should review those tests also to make sure you agree with their assertions.

coltfred (approved):

PR #372, opened 2026-03-12
Fix cache so that truly offline initialization with it works

skeet70:
This will require caches to be regenerated so it's "breaking", but only breaking in that the existing caches are incorrect, so it'll be a patch bump.


coltfred (approved):
LGTM

BobWall23 (approved):

PR #373, opened 2026-03-16
Add file encrypt and decrypt functions.

### LINKED ISSUE #358: Encrypt/decrypt for files to support large payloads
When encrypting or decrypting something potentially larger than memory, allow for passing in src and destination paths. Internally we should stream.

In the case of decrypt we'll stream to the file and if GCM fails, delete it.

skeet70:
Fixes #358 

Adds file encrypt and decrypt with constant memory use via internal streaming. This should be a swig-friendly API (unlike actual iterators/streams or an update/finalize into ?? setup).

See `src/crypto/streaming.rs` for the critical core bits and tests, along with a nice bit of ASCII art trying to explain what's going on. `encrypt_stream` and `decrypt_stream` in there are the next layer up and handle chunking. The next layer up is in `src/internal/document_api/file_ops.rs`, which deals with the ironcore header-y bits around the ciphertext, then `src/document/file.rs` which presents the final API for users (both managed and unmanaged).

### DIFF
--- a/src/crypto/streaming.rs
+++ b/src/crypto/streaming.rs
@@ -1,0 +1,1060 @@
+//! Streaming AES-GCM encryption/decryption using AES-CTR + incremental GHASH
+//!
+//! This module implements streaming AES-GCM by decomposing the algorithm into its two
+//! incremental components:
+//! - AES-CTR: Stream cipher for encryption/decryption (via aws-lc-rs)
+//! - GHASH: Authentication tag computation (via RustCrypto/ghash crate)
+//!
+//! This allows processing large files with constant memory usage while producing
+//! output compatible with standard AES-GCM (same format as `crypto::aes` module).
+//!
+// In Java/Scala/nodejs we were able to find common crypto libraries that implemented this model but did not find any
+// in Rust. If one is found (RustCrypto, ring, aws-lc-rs, etc), use that instead of doing it ourselves.
+
+use aws_lc_rs::{
+    cipher::{
+        AES_256, DecryptionContext, EncryptingKey, EncryptionContext, StreamingDecryptingKey,
+        StreamingEncryptingKey, UnboundCipherKey,
+    },
+    constant_time,
+};
+use ghash::{
+    GHash,
+    universal_hash::{KeyInit, UniversalHash},
+};
+use rand::{CryptoRng, RngCore};
+use std::{
+    io::{BufReader, BufWriter, Read, Write},
+    ops::DerefMut,
+    sync::Mutex,
+};
+
+use crate::{
+    IronOxideErr, Result,
+    crypto::aes::{AES_BLOCK_SIZE, AES_GCM_TAG_LEN, AES_IV_LEN, AES_KEY_LEN},
+    internal::take_lock,
+};
+
+/// Default block/chunk size for the file input/output (64 KB)
+pub(crate) const DEFAULT_IO_BLOCK_SIZE: usize = 64 * 1024;
+
+/// Build a counter block from IV and counter value.
+/// For 96-bit IVs: result = IV || counter (big-endian u32)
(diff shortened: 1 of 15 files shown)


zmre (approved):
Went over this pretty carefully and I think it's in good shape.

src/internal.rs
```diff
@@ -208,6 +212,12 @@ quick_error! {
         CacheSerdeError(error: postcard::Error) {
             source(error)
         }
+        AesGcmDecryptError {
+            display("AES-GCM decryption failed: authentication tag verification failed")
+        }
+        FileIOError { path: String, operation: String, message: String } {
+            display("File I/O error for '{}' during {}: {}", path, operation, message)
+        }
```
giarc3: This is called with an empty `path` a lot of times. I think we should probably make the path optional and have a slightly different message if it doesn't have a path, just so it's a little less confusing.

src/crypto/streaming.rs
```diff
@@ -0,0 +1,1183 @@
...
+        let complete_len = complete_blocks * AES_BLOCK_SIZE;
+
+        for chunk in self.pending[..complete_len].chunks_exact(AES_BLOCK_SIZE) {
+            let mut block = ghash::Block::default();
+            block.copy_from_slice(chunk);
+            self.ghash.update(&[block]);
+        }
+
+        // Keep only the remaining partial block
+        self.pending = self.pending[complete_len..].to_vec();
```
giarc3: I think you could use `drain` to avoid allocating a new vec every time

src/internal/document_api/file_ops.rs
```diff
@@ -0,0 +1,783 @@
...
+
+    // Stream encrypt the file content (writes IV + ciphertext + auth tag)
+    let key_bytes = *doc_sym_key.bytes();
+    streaming::encrypt_stream(&key_bytes, rng, &mut reader, &mut writer)?;
+    reset_file_permissions(destination_path)?;
+
+    // Encrypt DEK to all grantees
+    let recryption_result =
+        recrypt_document(&auth.signing_private_key, recrypt, dek, &doc_id, grants)?;
+    let encryption_errs = recryption_result.encryption_errs.clone();
```
giarc3: Why does this clone exist?

src/internal/document_api/file_ops.rs
```diff
@@ -0,0 +1,783 @@
...
+        .into_iter()
+        .chain(r.encryption_errs.clone())
+        .collect();
+
+    // Reset file permissions to normal (0644 on Unix)
+    reset_file_permissions(destination_path)?;
+
+    Ok(DocumentFileEncryptUnmanagedResult {
+        id: doc_id,
+
```
giarc3: empty line

giarc3 (approved):
Looks good, just a few suggestions

src/internal/document_api/file_ops.rs
skeet70: Vestigial, it used to be passed into a helper there.

coltfred (commented):
I didn't do too in depth of a review because of current time constraints, but if it were me I'd add some proptests around the update and accumulator.

For example a test that feeds the data all in at once vs feeding it in piece by piece in small chunks to ensure you keep the data held aside correctly. You can get a random bytes vec to encrypt and random places in that array to break it up.

I would also add a proptest for our streaming encrypt and the normal decrypt and normal encrypt with streaming decrypt just to better cover that base.

I did look at the presented API as well and it looks good. I'm going to assume you pulled it through to swig to verify that it presents well in the other langs.

skeet70:
@coltfred I had actually thought of adding proptests, but seeing as we didn't have any in the repo, didn't know if this rose to the level to bring them in. I'll add them, and probably add some to cover other things as well.

> I'm going to assume you pulled it through to swig to verify that it presents well in the other langs.

I didn't, I saw no reason it wouldn't. I will before release. (edit: it does come through fine)

PR #377, opened 2026-04-28
Add user_update_status and user_disable_self

giarc3:
- Add `user_disable_self` which a user can call to disable their account. Disabled users can still be members of groups but will be unable to call SDK functions.
  - Add `user_update_status` which uses a JWT to enable or disable a user.


BobWall23 (approved):

src/internal/user_api/requests.rs
```diff
@@ -289,24 +293,34 @@ pub mod user_update {
...
-        auth.request
-            .put(
-                &format!("users/{}", rest::url_encode(user_id.id())),
-                &req_body,
-                RequestErrorCode::UserUpdate,
-                AuthV2Builder::new(auth, OffsetDateTime::now_utc()),
-            )
-            .await
+        if status == Some(UserStatus::Enabled) {
+            Err(IronOxideErr::ValidationError(
```
coltfred: I don't think we should have this here. This is the request layer which isn't a place for validation. The server will reject this anyway and it's never called with Enabled.

If you're wanting to convey that, we should add a doc comment or write an integration test.

src/internal/user_api.rs
```diff
@@ -116,10 +116,50 @@ pub struct UserCreateResult {
...
+        serializer.serialize_u8(v)
+    }
+}
+
+impl<'de> serde::Deserialize<'de> for UserStatus {
+    fn deserialize<D: serde::Deserializer<'de>>(
+        deserializer: D,
+    ) -> std::result::Result<Self, D::Error> {
+        let v = u8::deserialize(deserializer)?;
+        match v {
```
coltfred, line 151: I wonder if deserialize needs to be a bit more permissive so backward compatibility can continue if we expand the server statuses. I looked for other places we're doing that here, but group status isn't exposed.

We could also separate the UserStatus public enum from the wire format.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 4.4.0
```
coltfred, line 3: Changing the status from usize to an enum would be breaking right?

coltfred (changes requested):
I played around with it a bit locally... I'm not sure the best path forward for the status enum. Fixing it here on the client will be a potential problem if we add another. 

Let me know your thoughts.

src/internal/user_api.rs
giarc3: Added Unknown(u8) and non_exhaustive to account for this

src/internal/user_api/requests.rs
giarc3: Removed

CHANGELOG.md
giarc3: Status was never exposed publicly before this

coltfred (approved):

PR #382, opened 2026-08-27
Webpki roots certs trust anchor

skeet70:
This fixes a downstream issue with `ironoxide-android`, whose cert verifier used OCSP (which is no longer supported by any of our cert providers).

Android now uses webpki-roots for trust anchors instead of the platform, the consequence of which is that `ironoxide-android` won't pick up platform trust-store CAs/custom CAs. If that's a problem for anyone before https://github.com/rustls/rustls-platform-verifier/issues/221 we may be able to work around it, but if not it'd be best to keep that workaround out.


src/internal/rest.rs
```diff
@@ -310,10 +310,56 @@ pub struct IronCoreRequest {
...
 
+/// Trust anchors for Android, taken from `webpki-roots` rather than the platform trust store.
+///
+/// reqwest 0.13's `rustls` feature verifies through `rustls-platform-verifier`, whose Android
+/// backend maps "certificate specifies no OCSP responder" onto `CertificateError::Revoked`
+/// (rustls-platform-verifier#221). Google Trust Services, Let's Encrypt and SSL.com have all
+/// stopped publishing OCSP responders, so that backend rejects chains that are valid and
+/// unrevoked. Other platforms keep the system trust store and the user-installed and enterprise
+/// CAs that come with it.
+#[cfg(all(feature = "tls-rustls", any(target_os = "android", test)))]
```
giarc3: so to be clear, all of our tests in this repo with rustls will now use webpki roots?

BobWall23 (approved):

src/internal/rest.rs
```diff
@@ -310,10 +310,56 @@ pub struct IronCoreRequest {
...
+    config.alpn_protocols = vec![b"h2".to_vec(), b"http/1.1".to_vec()];
+    config
+}
+
+#[cfg(all(feature = "tls-rustls", any(target_os = "android", test)))]
+lazy_static! {
+    /// Built once. Assembling the root store parses the entire `webpki-roots` bundle, and cloning
+    /// shares the resulting verifier behind an `Arc`.
+    static ref TLS_CONFIG: rustls::ClientConfig = webpki_roots_tls_config();
+}
```
giarc3: why is this one included in test if it's not used by the tests?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 4.4.2
+
+- On Android, the `tls-rustls` feature now trusts `webpki-roots` instead of the platform trust store. Android's verifier reports a certificate whose AIA omits an OCSP responder as revoked ([rustls-platform-verifier#221](https://github.com/rustls/rustls-platform-verifier/issues/221)), and the public CAs have stopped publishing OCSP responders. Android builds consequently no longer honor user-installed or enterprise CAs. Other platforms are unaffected.
+- `IronOxideErr::RequestError` now carries the underlying cause. reqwest's `Display` reports only the error kind and URL, which left a DNS failure, a refused connection, a rejected certificate and an elapsed timeout indistinguishable.
```
giarc3:
```suggestion
- `IronOxideErr::RequestError` now carries the underlying cause.
```

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
...
+- The Android rustls configuration is built once per process rather than per request.
```
giarc3:
```suggestion
```

giarc3 (approved):

src/internal/rest.rs
skeet70: No, because `webpki_roots_tls_config` still isn't used by the core code, just this function becomes available in tests so we can make some assertions about it without an Android emulator. I think only a small part of it is actually valuable outside of an emulator though, so I'll tighten it a bit.

src/internal/rest.rs
skeet70: It is used by the test on it specifically `webpki_roots_tls_config_advertises_http2_and_loads_roots`, not by the host tests generally (since they get the normal client).

I collapsed the function into this spot, and made the test only catch the part it really cares about, but this is still the case (that this is compiled for the tests).

